PR #219, opened 2015-02-02
Add rollbackWith for use together with retryWith

junjihashimoto:
Add rollbackWith for use together with retryWith.
When example(second argument) is failed, rollback(first argument) is used.
(When example is passed, rollback is not used.)


junjihashimoto:
Is this almost the same as bracketOnError of lifted-base?
This may be unnecessary.

PR #221, opened 2015-02-08
hspec-discover: Refactor slightly and add a help message

yamadapc:
This adds support for filtering the `spec` file paths using a regular
expression, through the `--regex` and `--not-regex` flags.

It's useful for (for example) dealing with temporary files from Emacs'
`flycheck-mode`, which can cause `hspec-discover` to trigger the
compilation of invalid modules (the temporary files themselves).

It also adds a help flag/message and refactors a couple of small parts
on the way.

**UPDATE 09/02/2015**
Okay; so the invalid module name rejection part was moved to #222. This now only contains the refactoring and help message addition.

### DIFF
--- a/hspec-discover/src/Config.hs
+++ b/hspec-discover/src/Config.hs
@@ -1,6 +1,7 @@
 module Config (
   Config (..)
 , defaultConfig
+, helpMessage
 , parseConfig
 , usage
 ) where
@@ -13,10 +14,11 @@ data Config = Config {
 , configFormatter :: Maybe String
 , configNoMain :: Bool
 , configModuleName :: Maybe String
+, configHelp :: Bool
 } deriving (Eq, Show)
 
 defaultConfig :: Config
-defaultConfig = Config False Nothing False Nothing
+defaultConfig = Config False Nothing False Nothing False
 
 options :: [OptDescr (Config -> Config)]
 options = [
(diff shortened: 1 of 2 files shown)


sol:
@yamadapc Nice catch, didn't know there was an issue with flycheck.  We definitely want to address this.

I think `hspec-discover` should always ignore `flycheck_*` files, not only if the user explicitly say so.  Otherwise you would need to modify `test/Spec.hs` of every project you want to contribute to.

In addition, I'm not super fond of the `regex-posix` dependency (which pulls in `containers` and `mtl`).

`flycheck_Foo` is not a valid module name, so it should always be safe to ignore `flycheck_FooSpec.hs`.  I think we can address this by ignoring all `*Spec.hs`/`*Spec.lhs` files that do not start with a valid module name.

@yamadapc would you be up for giving that a shot?  BTW, if you put refactorings and other improvements into separate PRs, then it's easier to review for me.

yamadapc:
I'm going to do that tonight. I'll probably make this PR contain just the `help` message and refactoring and make another one with the invalid module name filtering.

yamadapc:
@sol, just making sure you got a notification for this

sol:
I think a proper help message for `hspec-discover` is a "nice to have", but  not super crucial.  It is a pre-processor and not typically invoked directly by users.  But it could even be useful to make that faked clear in a help message.

If we add help messages, then we should do it TheRightWay™, which is more complicated then I feel it should be.  I'll add some comments.

hspec-discover/src/Config.hs
```diff
@@ -24,6 +26,7 @@ options = [
   , Option [] ["formatter"] (ReqArg (\s c -> c {configFormatter = Just s}) "FORMATTER") ""
   , Option [] ["module-name"] (ReqArg (\s c -> c {configModuleName = Just s}) "NAME") ""
   , Option [] ["no-main"] (NoArg $ \c   -> c {configNoMain = True}) ""
+  , Option "h" ["help"] (NoArg $ \c -> c {configHelp = True}) ""
```
sol: Following GNU conventions (e.g. check `cp`), I would only add `--help`, but would skip the `-h` shortcut.

hspec-discover/src/Config.hs
```diff
@@ -32,14 +35,17 @@ usage prog = "\nUsage: " ++ prog ++ " SRC CUR DST [--module-name=NAME]\n"
...
-           else
-             Right c
+        c = foldl (flip id) defaultConfig opts
+      in validateConfig c
     (_, _, err:_)  -> formatError err
     (_, arg:_, _)  -> formatError ("unexpected argument `" ++ arg ++ "'\n")
   where
     formatError err = Left (prog ++ ": " ++ err ++ usage prog)
+
+    validateConfig c | configHelp c = Left (helpMessage prog)
```
sol: Help messages should go to `stdout` and the program should exit with exit code `0`.  Returning a `Left` here will result in help messages being treated the same way as errors.  They will be printed to `stderr` and the program will terminate with `exitFailure`.

hspec-discover/src/Run.hs
```diff
@@ -41,16 +41,18 @@ run args_ = do
...
       Right conf -> do
         when (configNested conf) (hPutStrLn stderr "hspec-discover: WARNING - The `--nested' flag is deprecated and will be removed in a future release!")
         specs <- findSpecs src
         writeFile dst (mkSpecModule src conf specs)
-    _ -> do
-      hPutStrLn stderr (usage name)
-      exitFailure
+    _ -> exitError $ if "-h" `elem` args_ || "--help" `elem` args_
+            then helpMessage name
+            else usage name
```
sol: I would get rid of that case and instead always print something like

```
hspec-discover: unrecognized option `--foo'
Try `hspec-discover --help' for more information.
```

if the user specifies an invalid argument.  This is consistent with what `hspec` and most other tools do and gives the user the chance to discover `--help` himself.

sol:
You could also look at the command-line handling of hspec (in https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Options.hs), but it's messy, and I would love to extract it into a library some time, as I have written code along that lines several times already in the past.

It's still up to you if you want to go for it.  But if you do so, I'm happy to provide feedback.  Just ping me.

yamadapc:
I've refactored to address your concerns. I'm not sure what you'd want to refactor into a library, just the `help` message handling? Or command-line option parsing, in lieu of `docopt`. I'd love to help with whatever is possible, if you want to delegate some things :)

sol:
@yamadapc sorry for sitting on this patch for so long.

`--no-main` and `--formatter` are deprecated. As for `--nested` I'm not sure yet what its fate will be. For that reason I prefer to let those things undocumented and stick with the current minimalistic help we have.

PR #222, opened 2015-02-09
hspec-discover: Reject invalid module names

yamadapc:
Tweaks `fileToSpec` so that it rejects files with invalid module names,
as discussed in #221. Please let me know if you'd like me to test or
change anything. I know I shouldn't have refactored the function as well
as adding this feature, but I thought it looked a little clearer like
this.

Let me know what you think.

### DIFF
--- a/hspec-discover/src/Run.hs
+++ b/hspec-discover/src/Run.hs
@@ -121,12 +121,21 @@ findSpecs src = do
 fileToSpec :: FilePath -> FilePath -> Maybe Spec
 fileToSpec dir file = case reverse $ splitDirectories file of
   x:xs -> case stripSuffix "Spec.hs" x <|> stripSuffix "Spec.lhs" x of
-    Just name | (not . null) name -> Just . Spec (dir </> file) $ (intercalate "." . reverse) (name : xs)
+    Just name | isValidModuleName name && all isValidModuleName xs ->
+      Just . Spec (dir </> file) $ (intercalate "." . reverse) (name : xs)
     _ -> Nothing
   _ -> Nothing
   where
     stripSuffix :: Eq a => [a] -> [a] -> Maybe [a]
     stripSuffix suffix str = reverse <$> stripPrefix (reverse suffix) (reverse str)
+
+-- See `Cabal.Distribution.ModuleName` (http://git.io/bj34)
+isValidModuleName :: String -> Bool
+isValidModuleName [] = False
+isValidModuleName (c:cs) = isUpper c && all isValidModuleChar cs
+
+isValidModuleChar :: Char -> Bool
+isValidModuleChar c = isAlphaNum c || c == '_' || c == '\''
 
 getFilesRecursive :: FilePath -> IO [FilePath]
 getFilesRecursive baseDir = sort <$> go []
--- a/hspec-discover/test/RunSpec.hs
+++ b/hspec-discover/test/RunSpec.hs
@@ -84,6 +84,14 @@ spec = do
     it "returns Nothing for invalid spec name" $ do
       fileToSpec "" "foo" `shouldBe` Nothing
 
+    context "when spec does not have a valid module name" $ do
+      it "returns Nothing" $ do
+        fileToSpec "" "flycheck_Spec.hs" `shouldBe` Nothing
+
+    context "when any component of a hierarchical module name is not valid"$ do
+      it "returns Nothing" $ do
+        fileToSpec "" ("Valid" </> "invalid"  </>"MiddleNamesSpec.hs") `shouldBe` Nothing
+
     context "when path has directory component" $ do
       it "converts path to spec name" $ do
         let file = "Foo" </> "Bar" </> "BazSpec.hs"
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -1,3 +1,7 @@
+## Changes in 2.1.4
+  - Make `hspec-discover` ignore modules with invalid module names, this fixes
+    issues with `flycheck`'s temporary files
+
 ## Changes in 2.1.3
   - Format source locations like gcc does
 
--- a/README.markdown
+++ b/README.markdown
@@ -65,3 +65,4 @@ To run the test suite do:
  * Junji Hashimoto
  * Jan Matějka
  * Ömer Sinan Ağacan
+ * Pedro Tacla Yamada


yamadapc:
btw, I see there's a `travis-ci` configuration for this project, but that it isn't running on PRs. Is that on purpose?

sol:
@yamadapc Sorry for the late reply.  Not sure why travis won't build PRs.  I would assumed it worked in the past.

hspec-discover/src/Run.hs
```diff
@@ -119,15 +119,31 @@ findSpecs src = do
   mapMaybe (fileToSpec dir) . filter (/= file) <$> getFilesRecursive dir
 
 fileToSpec :: FilePath -> FilePath -> Maybe Spec
-fileToSpec dir file = case reverse $ splitDirectories file of
-  x:xs -> case stripSuffix "Spec.hs" x <|> stripSuffix "Spec.lhs" x of
-    Just name | (not . null) name -> Just . Spec (dir </> file) $ (intercalate "." . reverse) (name : xs)
-    _ -> Nothing
-  _ -> Nothing
+fileToSpec dir sfile = do
+    file@(_:_) <- stripSuffix "Spec.hs" sfile <|>
```
sol: Checking for `null` like his works, but I'm not sure if I like it better than an explicit guard.

hspec-discover/src/Run.hs
```diff
@@ -119,15 +119,31 @@ findSpecs src = do
...
+                  stripSuffix "Spec.lhs" sfile
+    mname <- getModuleName file
+    return $ Spec (dir </> sfile) mname
   where
     stripSuffix :: Eq a => [a] -> [a] -> Maybe [a]
     stripSuffix suffix str = reverse <$> stripPrefix (reverse suffix) (reverse str)
 
+getModuleName :: FilePath -> Maybe String
+getModuleName f
+    | last f == pathSeparator = Nothing
```
sol: `last` is partial.  Even though we checked for `null` above, I still think a solution that uses pattern matching is preferable.

hspec-discover/src/Run.hs
```diff
@@ -119,15 +119,31 @@ findSpecs src = do
   mapMaybe (fileToSpec dir) . filter (/= file) <$> getFilesRecursive dir
 
 fileToSpec :: FilePath -> FilePath -> Maybe Spec
-fileToSpec dir file = case reverse $ splitDirectories file of
-  x:xs -> case stripSuffix "Spec.hs" x <|> stripSuffix "Spec.lhs" x of
-    Just name | (not . null) name -> Just . Spec (dir </> file) $ (intercalate "." . reverse) (name : xs)
```
sol: We could just replace `(not . null) name` with `isValidModuleName name && all isValidModuleName xs`.  I would prefer that not only because it's a smaller change, but also for the reasons given in my other comments.

hspec-discover/test/RunSpec.hs
```diff
@@ -83,6 +83,12 @@ spec = do
 
     it "returns Nothing for invalid spec name" $ do
       fileToSpec "" "foo" `shouldBe` Nothing
+      fileToSpec "" "fooSpec.hs" `shouldBe` Nothing
+      fileToSpec "" "flycheck_Spec.hs" `shouldBe` Nothing
+      fileToSpec "" ("Valid" </> "invalid"  </>"MiddleNamesSpec.hs")
+          `shouldBe` Nothing
+      fileToSpec "" "Weird.Spec.hs" `shouldBe` Nothing
+      fileToSpec "/" "Spec.hs" `shouldBe` Nothing
```
sol: Let's stick with one expectation per test.  I would be happy with two cases here:

```
context "when spec does not have a valid module name" $ do
  it "returns Nothing" $ do
    fileToSpec "" "flycheck_Spec.hs" `shouldBe` Nothing

context "when any component of a hierarchical module name is not valid" $ do
  it "returns Nothing" $ do
    fileToSpec "" ("Valid" </> "invalid" </>"MiddleNamesSpec.hs") `shouldBe` Nothing
```

If we want to tests for a wide variety of invalid module names, then that should be tests for `isValidModuleName`.  But I would not do so, as we copy & pasted `isValidModuleName` anyway, and I mostly write tests to drive the implementation.

yamadapc:
I like that this is a smaller change, but I thought the changes to `fileToSpec` made it a lot cleaner. Either way, I've rebased to fix the issues you pointed out.

What I was thinking was:
```haskell
fileToSpec dir sfile = do
  file <- stripSuffix "Spec.hs" sfile <|> stripSuffix "Spec.lhs" sfile
  Spec (dir </> file) <$> getModuleName file
  where -- ...
```

Adding the `[]` pattern to `getModuleName` would make this a very clean solution.

hspec-discover/test/RunSpec.hs
```diff
@@ -84,6 +84,14 @@ spec = do
     it "returns Nothing for invalid spec name" $ do
       fileToSpec "" "foo" `shouldBe` Nothing
 
+    context "when spec does not have a valid module name" $ do
+        it "returns Nothing" $ do
+            fileToSpec "" "flycheck_Spec.hs" `shouldBe` Nothing
```
sol: I think both 4 spaces and 2 spaces are valid choices for indentation, but for consistency with the rest of the code, let's stick to 2 spaces here ;)

hspec-discover/test/RunSpec.hs
```diff
@@ -84,6 +84,14 @@ spec = do
...
+
+    context "when any component of a hierarchical module name is not valid"$ do
+        it "returns Nothing" $ do
+            fileToSpec "" ("Valid" </> "invalid"  </>"MiddleNamesSpec.hs") `shouldBe` Nothing
```
sol: And here ;)

sol:
@yamadapc This looks good now.  Please add yourself to the list of contributors in `README.md` + update `CHANGES.markdown` (new version will be `2.1.4`).  I'll then merge and release to Hackage.

Regarding your suggested refactoring, if you think there is something to gain here, you can open a separate PR if you want.  But let's first get this out, as I think it's a substantial enhancement for all flycheck users.

sol:
Also, please squash all commits into a single commit if possible.

yamadapc:
I think things are shining now. Do you want me to do something to the `help` message PR, or do you think it's not worth having it merged?

sol:
@yamadapc That looks great now, thanks a lot!

sol:
released to haddock

PR #234, opened 2015-05-11
Revert "Add captureLogs function"

jkarni:
This reverts commit 3bf50f54e51aa9d2f606098d0bf73582df2dac4a.


jkarni:
Moved to [mocker](https://github.com/hspec/mockery/pull/2/files)

hspec-contrib/src/Test/Hspec/Contrib/LoggingFacade.hs
```diff
@@ -1,25 +0,0 @@
-module Test.Hspec.Contrib.LoggingFacade (
-  captureLogs
-) where
```
sol: This should be deprecated only, with a reference to `mockery`.

PR #235, opened 2015-05-11
Deprecate captureLogs

jkarni:


jkarni:
See https://github.com/hspec/mockery/pull/2/files

sol:
@jkarni I ended up moving the warning to the module, so that even an unused import will generate a warning.

sol:
https://github.com/hspec/hspec/commit/f30eb0ece13d38c85c0c57eab6d0713928f16ed8

sol:
On hackage.

PR #239, opened 2015-07-06
Precisely pinpoint ‘it’ locations using CallStack (proof of concept)

Fuuzetsu:
It actually works so well that it breaks the discover tests. I don't
‘fix’ them here because it's not clear how it should be done.

```
*Test.Hspec.DiscoverSpec> main

postProcessSpec
  adds heuristic source locations FAILED [1]
  when same requirement is used multiple times
    assigns locations sequentially FAILED [2]
    when a requirement occurs more often in the spec tree than in the source file
      assigns Nothing FAILED [3]

Failures:

  1) postProcessSpec adds heuristic source locations
       expected: Just Location {locationFile = "/run/user/1000/temp7198853861649760492", locationLine = 2, locationColumn = 0, locationAccuracy = BestEffort}
        but got: Just Location {locationFile = "test/Test/Hspec/DiscoverSpec.hs", locationLine = 34, locationColumn = 11, locationAccuracy = ExactLocation}

  2) postProcessSpec, when same requirement is used multiple times, assigns locations sequentially
       expected: Just Location {locationFile = "/run/user/1000/temp5965166491189641421", locationLine = 1, locationColumn = 0, locationAccuracy = BestEffort}
        but got: Just Location {locationFile = "test/Test/Hspec/DiscoverSpec.hs", locationLine = 49, locationColumn = 13, locationAccuracy = ExactLocation}

  3) postProcessSpec, when same requirement is used multiple times, when a requirement occurs more often in the spec tree than in the source file, assigns Nothing
       expected: Nothing
        but got: Just Location {locationFile = "test/Test/Hspec/DiscoverSpec.hs", locationLine = 63, locationColumn = 15, locationAccuracy = ExactLocation}
```

--

Using GHC HEAD (20150705)


sol:
This looks awesome! :)

Regarding the failing tests, they should be excluded with `CPP` when we are building with a version of GHC that gives us exact locations.

hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -44,8 +51,23 @@ describe label spec = runIO (runSpecM spec) >>= fromSpecList . return . specGrou
...
+    -- The IP type signatures on ?loc here and on ‘it’ itself as well
+    -- as the actual IP names are *essential*: if the names differ or
+    -- one of the signatures is missing, IP scoping rules or CallStack
+    -- rules will mess the stack up and we won't get what we wanted.
+    srcLoc = case reverse (getCallStack (?loc :: (?loc :: CallStack) => CallStack)) of
+      ("it", loc):_ -> Just loc
+      _ -> Nothing
+
+    toLocation cwd s =
+      Location (makeRelative cwd $ srcLocFile s) (srcLocStartLine s) (srcLocStartCol s) ExactLocation
```
sol: How exactly does the path look if you don't apply `makeRelative` here?  Note that the `cwd` may be different from the source code location at the time the test suite is run.  If we can decide on the final path statically rather than doing it at run time, I would prefer that.

Fuuzetsu: It's absolute path, all the way from root.

sol: Using the binary from @nh2 (https://github.com/nh2/ghc/releases/tag/ghc-7.10.1-release-srcloc-ip), I get relative paths.  From my perspective it's desirable for paths in source to follow the exact same semantics as the ones in regular compile error messages.

@Fuuzetsu can you try to find out whether the code changed in the meanwhile, or is it due to your build environment?  E.g. can you compare the format of compile errors and src locations in your environment?

Fuuzetsu:
Just for the record, after some mails to `ghc-devs`, this feature did make it into 7.10.2. 7.10.2 seems to have been cut about 12 hours ago on the branch so should be officially out soon.

sol:
That's exciting news!

sol:
On master (see #240).

PR #241, opened 2015-07-26
Add -j/--jobs and --unlimited-jobs to allow users to limit parallelism

ryantrinkle:
For some jobs, particularly when running in an environment like travis-ci, running a huge number of parallel tests can cause memory exhaustion and/or poor performance.  This patch allows hspec to be invoked with a -j or --jobs option to limit this parallelism.  The current default behavior, allowing unlimited jobs in parallel, remains the default.  An --unlimited-jobs switch to explicitly invoke the default behavior is also included for completeness.

### DIFF
--- a/doc/parallel-spec-execution.md
+++ b/doc/parallel-spec-execution.md
@@ -45,4 +45,7 @@ All specs that are not explicitly marked with `parallel` are run in the
 application's main thread.
 
 
-By default, the number of threads available for parallel execution is equal to the number of processors available to the process, as determined by `+RTS -N<n>`.  A different number (higher or lower) can be specified using the `-j` flag.  Note that this number is in addition to the main thread, and must be at least 1 (otherwise, parallelizable jobs would not be able to run).
+By default, the number of threads available for parallel execution is equal to
+the number of processors available to the process, as determined by
+`+RTS -N<n>`.  A different number (higher or lower) can be specified using the
+`-j` flag.  Note that this number is in addition to the main thread.
--- a/doc/parallel-spec-execution.md
+++ b/doc/parallel-spec-execution.md
@@ -43,3 +43,6 @@ Link the program with `-threaded` and pass `+RTS -N -RTS` when running it:
 
 All specs that are not explicitly marked with `parallel` are run in the
 application's main thread.
+
+
+By default, the number of threads available for parallel execution is equal to the number of processors available to the process, as determined by `+RTS -N<n>`.  A different number (higher or lower) can be specified using the `-j` flag.  Note that this number is in addition to the main thread, and must be at least 1 (otherwise, parallelizable jobs would not be able to run).
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -25,6 +25,7 @@ import           System.IO
 import           System.Environment (getProgName, getArgs, withArgs)
 import           System.Exit
 import qualified Control.Exception as E
+import           Control.Concurrent
 
 import           System.Console.ANSI (hHideCursor, hShowCursor)
 import qualified Test.QuickCheck as QC
@@ -122,13 +123,17 @@ hspecWithResult conf spec = do
         seed = (fromJust . configQuickCheckSeed) c
         qcArgs = configQuickCheckArgs c
 
+    jobsSem <- case configMaxParallelJobs c of
+      Nothing -> return Nothing
+      Just maxJobs -> liftM Just $ newQSem maxJobs
+
     useColor <- doesUseColor h c
 
     filteredSpec <- filterSpecs c . applyDryRun c <$> runSpecM spec
 
     withHiddenCursor useColor h $
       runFormatM useColor (configHtmlOutput c) (configPrintCpuTime c) seed h $ do
-        runFormatter useColor h c formatter filteredSpec `finally_` do
+        runFormatter jobsSem useColor h c formatter filteredSpec `finally_` do
           failedFormatter formatter
 
         footerFormatter formatter
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -123,9 +123,9 @@ hspecWithResult conf spec = do
         seed = (fromJust . configQuickCheckSeed) c
         qcArgs = configQuickCheckArgs c
 
-    jobsSem <- case configMaxParallelJobs c of
-      Nothing -> return Nothing
-      Just maxJobs -> liftM Just $ newQSem maxJobs
+    jobsSem <- newQSem =<< case configMaxParallelJobs c of
+      Nothing -> getNumCapabilities
+      Just maxJobs -> return maxJobs
 
     useColor <- doesUseColor h c
 
--- a/hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
@@ -22,8 +22,8 @@ import           Test.Hspec.Timer
 type EvalTree = Tree (ActionWith ()) (String, Maybe Location, ProgressCallback -> FormatResult -> IO (FormatM ()))
 
 -- | Evaluate all examples of a given spec and produce a report.
-runFormatter :: Bool -> Handle -> Config -> Formatter -> [SpecTree ()] -> FormatM ()
-runFormatter useColor h c formatter specs = do
+runFormatter :: Maybe QSem -> Bool -> Handle -> Config -> Formatter -> [SpecTree ()] -> FormatM ()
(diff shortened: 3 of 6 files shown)


sol:
@ryantrinkle Sorry for being lazy on this one.  I'm actually puzzled whether we want to limit concurrency in general?  @soenkehahn can we take a look tomorrow?

ryantrinkle:
The thing that prompted me to do this was that I had a test suite that I was running on travis-ci, and it was getting killed for exhausting the available memory, due to running *all* of its tests in parallel.  By limiting to, say, 2 or 3 tests simultaneously, I can prevent it from using so much RAM.

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -122,13 +123,17 @@ hspecWithResult conf spec = do
         seed = (fromJust . configQuickCheckSeed) c
         qcArgs = configQuickCheckArgs c
 
+    jobsSem <- case configMaxParallelJobs c of
+      Nothing -> return Nothing
```
soenkehahn: In case `-j` is omitted can we use `getNumCapabilities` as a default? That would mean that we *always* use a semaphore, sometimes with a capacity of 1. This should hopefully simplify the code.

hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
```diff
@@ -53,22 +53,27 @@ every seconds action = do
...
-runSequentially e reportProgress formatResult = return $ do
-  result <- liftIO $ evalExample (e reportProgress)
+limitParallelism :: Maybe QSem -> IO a -> IO a
+limitParallelism jobsSem = case jobsSem of
+  Nothing -> id
+  Just s -> E.bracket_ (waitQSem s) (signalQSem s)
+
+runSequentially :: Maybe QSem -> (ProgressCallback -> IO Result) -> ProgressCallback -> FormatResult -> IO (FormatM ())
+runSequentially jobsSem e reportProgress formatResult = return $ do
+  result <- liftIO $ limitParallelism jobsSem $ evalExample (e reportProgress)
```
soenkehahn: This is a little bit tough to explain, but we (@sol and I) were just discussing parallel execution of specs. I think that using `limitParallelism` here can result in all parallelizable examples to be executed first while the sequential ones are blocked. After the parallelizable examples are done the sequential ones are being executed in sequence. This is less than ideal parallelism. Ideally we would continue to run the sequential examples in the main thread while the parallelizable ones are also being run.

Maybe it's best to just remove `limitParallelism` here.

(Strictly speaking we may get one more thread than specified with `-j` in the presence of sequential examples. But we assume that's ok.)

hspec-core/src/Test/Hspec/Options.hs
```diff
@@ -137,6 +139,8 @@ options = [
   , Option   []  ["dry-run"]          (NoArg setDryRun)                   (h "pretend that everything passed; don't verify anything")
   , Option   []  ["fail-fast"]        (NoArg setFastFail)                 (h "abort on first failure")
   , Option   "r" ["rerun"]            (NoArg  setRerun)                   (h "rerun all examples that failed in the previously test run (only works in GHCi)")
+  , mkOption "j"  "jobs"              (Arg "N" readMaybe setMaxJobs)      (h "run at most N tests simultaneously (default: unlimited)")
+  , Option   []  ["unlimited-jobs"]   (NoArg clearMaxJobs)                (h "allow an unlimited number of jobs to run simultaneously (this is the default)")
```
soenkehahn: From our perspective let's just skip `unlimited-jobs`.

hspec-core/src/Test/Hspec/Options.hs
```diff
@@ -137,6 +139,8 @@ options = [
   , Option   []  ["dry-run"]          (NoArg setDryRun)                   (h "pretend that everything passed; don't verify anything")
   , Option   []  ["fail-fast"]        (NoArg setFastFail)                 (h "abort on first failure")
   , Option   "r" ["rerun"]            (NoArg  setRerun)                   (h "rerun all examples that failed in the previously test run (only works in GHCi)")
+  , mkOption "j"  "jobs"              (Arg "N" readMaybe setMaxJobs)      (h "run at most N tests simultaneously (default: unlimited)")
```
soenkehahn: The help text for the default value has to be updated. (See above.)

hspec-core/src/Test/Hspec/Options.hs
```diff
@@ -150,12 +154,18 @@ options = [
     setOutputFile :: String -> Config -> Config
     setOutputFile file c = c {configOutputFile = Right file}
 
+    setMaxJobs :: Int -> Config -> Config
+    setMaxJobs n c = c {configMaxParallelJobs = Just n}
+
+
+
```
soenkehahn: Please, remove two empty lines.

soenkehahn:
@ryantrinkle: @sol and I looked at it and wrote some comments. In principle this looks like a very useful commit.

hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
ryantrinkle: Perhaps the right thing to do would be to reserve one job for the sequential operations by starting the semaphore with one less than the provided number, then "give it back" by signaling the semaphore once when the final sequential job is done.  I considered doing it this way, but I couldn't see where in the code to add that final signalQSem operation.  If you can point that out to me, I'll be happy to change the implementation to do it that way.

hspec-core/src/Test/Hspec/Core/Runner.hs
ryantrinkle: That makes a lot of sense to me.  There may be some workloads that are IO-bound rather than CPU-bound, so jobs in excess of getNumCapabilities could make sense.  However, the user could just select this manually if they need it.

sol: Yes, exactly same reasoning here.

hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
sol: I think that would deadlock for `-j 1`, no?

ryantrinkle: I don't think so.  The sempahore would have an initial value of 0, meaning that the parallelizable jobs would block.  However, the sequential jobs would proceed (as they would no longer wait on the semaphore at all), and, when they completed, the parallelizable jobs would then begin running.

sol: My reasoning is that the main thread, when producing the spec output, will eventually try to take the result of a parallelizable example from an `MVar` and block on that operation.  So the main thread would need to signal before it prints the result of the first parallelizable example, I guess.  But after the first parallelizable example there may still be more sequential examples, so proceeding naively we would end up with one more thread than specified again.

I imagine it may be possible to get it "right" by using wait/signal multiple times in the main thread, but I'm not sure the added complexity is worth it.

sol:
We also need to update doc/parallel-spec-execution.md

sol:
And yes, this is a crucial change. My original implementation was far too naive.

hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
ryantrinkle: Ah, OK.  I didn't realize that that kind of data dependency could exist.

ryantrinkle:
Hi guys,

I just pushed a commit that I believe addresses all of the concerns you raised.  In addition to the specific changes you requested, I also added validation to the -j parameter (to prevent users from using  -j 0 and receiving an unhelpful "blocked on MVar" message).  Also, since Hspec's own test suite runs with 1 capability the "hspecResult runs specs in parallel" test began failing; I fixed this by explicitly passing -j to that particular test.

doc/parallel-spec-execution.md
```diff
@@ -43,3 +43,6 @@ Link the program with `-threaded` and pass `+RTS -N -RTS` when running it:
 
 All specs that are not explicitly marked with `parallel` are run in the
 application's main thread.
+
+
+By default, the number of threads available for parallel execution is equal to the number of processors available to the process, as determined by `+RTS -N<n>`.  A different number (higher or lower) can be specified using the `-j` flag.  Note that this number is in addition to the main thread, and must be at least 1 (otherwise, parallelizable jobs would not be able to run).
```
soenkehahn: I think I would remove the note about the about `-j` having to be `>= 1`. If we have a proper warning we don't need a note here. (Maybe we should add a test for having a good warning.)

And this file seems to be word-wrapped.

soenkehahn:
Hi @ryantrinkle,

many thanks for addressing all our comments. Apart from my comment on the docs this LGTM.

soenkehahn:
@ryantrinkle: Many thanks for this!!!

PR #246, opened 2015-11-15
Make sure that Vim's default 'errorformat' recognizes exact locations

sol:


sol:
@soenkehahn Do you think this is the right thing to do?

soenkehahn:
Yes. I hate trailing slashes in program output, but if `vim`'s standard error format needs that, then I would do it.

PR #247, opened 2015-11-23
workaround for ghcjs

soenkehahn:
This is a workaround for https://github.com/ghcjs/ghcjs/issues/263

The test-suite of `hspec` currently doesn't run with `ghcjs` because the functions from `silently` also don't work with `ghcjs`. However I have been able to successfully run other `hspec` test-suites compiled with `ghcjs` with this branch.

Once `silently` works with `ghcjs` the tests for `FailureReport` have to be omitted with that compiler. I didn't want to include any `CPP` code for that yet, because the test-suite doesn't run anyways.

@sol: Are you still fine with merging this, pending travis?

(cc: @luite)

### DIFF
--- a/hspec-core/src/Test/Hspec/FailureReport.hs
+++ b/hspec-core/src/Test/Hspec/FailureReport.hs
@@ -1,13 +1,17 @@
+{-# LANGUAGE CPP #-}
 module Test.Hspec.FailureReport (
   FailureReport (..)
 , writeFailureReport
 , readFailureReport
 ) where
 
-import           System.IO
+#ifndef __GHCJS__
 import           System.SetEnv
+import           Test.Hspec.Core.Util (safeTry)
+#endif
+import           System.IO
 import           Test.Hspec.Compat
-import           Test.Hspec.Core.Util (Path, safeTry)
+import           Test.Hspec.Core.Util (Path)
 
 data FailureReport = FailureReport {
   failureReportSeed :: Integer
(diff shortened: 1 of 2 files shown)


hspec-core/src/Test/Hspec/FailureReport.hs
```diff
@@ -25,8 +32,12 @@ writeFailureReport x = do
   where
     onError err = do
       hPutStrLn stderr ("WARNING: Could not write environment variable HSPEC_FAILURES (" ++ show err ++ ")")
+#endif
 
 readFailureReport :: IO (Maybe FailureReport)
+#ifdef __GHCJS__
+readFailureReport = return Nothing
```
sol: Is this required? `lookupEnv` works with GHCJS, right?  So I guess this would result in `Nothing` anyway, even without the CPP.

hspec-core/src/Test/Hspec/FailureReport.hs
```diff
@@ -18,6 +22,9 @@ data FailureReport = FailureReport {
 } deriving (Eq, Show, Read)
 
 writeFailureReport :: FailureReport -> IO ()
+#ifdef __GHCJS__
+writeFailureReport _ = return ()
```
sol: Can we have a comment here, describing the reason for this + e.g. url to corresponding issues.

sol:
@soenkehahn Yes, I'm still fine with this change.

sol:
+ we want an entry in `CHANGES.markdown`

hspec-core/src/Test/Hspec/FailureReport.hs
soenkehahn: Yes, tested just now, `lookupEnv` works. I'll change this.

soenkehahn:
@sol: I've tried to address all your comments. I guess this should be `2.2.1` (and not `2.2.0.1`). Looking into travis now.

soenkehahn:
Travis is green now. Before it was just a network problem, it seems.

sol:
LGTM.

@soenkehahn do you want to cut a release?  I think you have upload rights to Hackage already.

soenkehahn:
Yes, I don't know when I'll get to it, though.

PR #251, opened 2016-01-11
Fix compilation for base-4.9

hvr:


hvr:
This seems all that's needed for `cabal install hspec` to succeed with GHC 8

sol:
@hvr sorry for the delay and thanks a lot for the patch.  I'll upload to Hackage today.

PR #261, opened 2016-04-19
Reading command line configuration options from files

### LINKED ISSUE #260: Add config file, e.g. ~/.hspec
... so that I can e.g. specify `--format=progress` globally ...

triplepointfive:
Reads options from global `~/.hspec` and local `./.hspec` files with following priority: global < local < direct args.
For invalid options it also prints the source of issue, like 

> unrecognized option `--foo' (defined in /Users/Smelkov/.hspec)

Some workaround with test env is required because hspec-core applied global options to itself while testing, but seems fine with hspec package though.

Resolves #260

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -31,6 +31,10 @@ import           Data.Maybe
 import           System.IO
 import           System.Environment (getProgName, getArgs, withArgs)
 import           System.Exit
+#ifndef TEST
+import           System.FilePath (combine, makeRelative)
+import           System.Directory (getHomeDirectory, doesFileExist)
+#endif
 import qualified Control.Exception as E
 import           Control.Concurrent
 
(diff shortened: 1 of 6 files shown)


sol:
Sorry for getting back late! I was off the scene for personal reasons.

Regarding your code, yes, we want that future, but as far as I can tell the majority of your code is not tested. Given this situation I am not willing to review this.

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -173,6 +175,34 @@ hspecWithResult conf spec = do
...
+    , ("", args)
+    ]
+
+-- | Reads command line configuration options from file.
+readConfFileOptions :: FilePath -> IO [String]
+readConfFileOptions fileName = do
+  confFileExists <- doesFileExist fileName
+  if confFileExists then lines <$> readFile fileName else return []
+
+#endif
```
sol: This whole chunk of code is not tested, right?

Issue #269, opened 2016-07-31
Send an exception to terminating threads


cocreature:
If you use `bracket` in one of your tests (`around` doesn’t make sense if you want it to be part of your test) you would expect that your cleanup action is always run, especially in the (at least for me common case) of issuing a user interrupt. However because `hspec` runs the tests in separate threads they are just killed and you end up leaking resources.

A simple solution would be to use `withAsync` instead of `forkIO` in `runParallel`. 

I’d be happy to provide a PR for that if you agree with these semantics.

cocreature:
It looks like this is not actually the problem in my case since I’m running in sequential mode.

I’ll leave this issue open nevertheless since this is still what I would consider a bug in parallel mode.

sol:
I believe this is fixed by ce5dce4289394a3b33922d695acc45ce60ea66ea.

Closing.

PR #270, opened 2016-08-02
Block until finalizers return

cocreature:
This was the bug I originally thought was caused by #269 (I am pretty sure there is still a separate issue when running in parallel).

The problem is that when the main thread exits all other threads are killed instantly without even being noticed about it. `withAsync` helps somewhat as it throws an exception to these threads. However it returns after the exception has been raised which does not mean that the finalizers have finished running. You thus run into cases where your finalizer is abruptly terminated somewhere in the middle. By waiting we make sure that all the finalizers have been run.

Obviously this means that we can deadlock if a finalizer never returns, but that is always the case for `bracket` and imho leaking resources (in my case I was leaking temporary databases) is a lot worse.

I made a [simple example](https://gist.github.com/cocreature/f7a6e1b08face3911222202ca84ae812) to illustrate the problem.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Util.hs
+++ b/hspec-core/src/Test/Hspec/Core/Util.hs
@@ -129,4 +129,13 @@ formatException err@(SomeException e) = case fromException err of
 -- occurs, the exception is returned instead.  Unlike `try` it is agnostic to
 -- asynchronous exceptions.
 safeTry :: IO a -> IO (Either SomeException a)
-safeTry action = withAsync (action >>= evaluate) waitCatch
+safeTry action =
+  bracket (async ((action >>= evaluate)))
+          -- It is important to wait here to make sure all finalizers
+          -- in action have been run. Otherwise the main thread can
+          -- exit before they have finished and the finalizers are
+          -- only partially run.
+          (\a -> cancel a >> waitCatch a) -- We use waitCatch to hide
+                                          -- the ThreadKilled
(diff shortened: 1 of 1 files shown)


sol:
@cocreature sorry for the delay on this, and thanks a lot for tracking this down!

I pushed a failing test case to `master` right now, and I will merge this later.

As a side note, as I understand it, this will help with `SIGINT`, but not with `SIGTERM`.

cocreature:
> As a side note, as I understand it, this will help with SIGINT, but not with SIGTERM.

You are right I wasn’t aware of that! I guess you are far more likely to send a SIGINT to your test suite than SIGTERM but it’s at least unexpected.

sol:
Yes, true.

sol:
Also pinging @snoyberg, who wrote that original piece code and a blog post on the issue of "catching almost all exceptions".

sol:
And for reference https://ghc.haskell.org/trac/ghc/ticket/10169

snoyberg:
I don't see anything wrong with the patch, but generally speaking relying on a finalizer to clean up resources outside of the process is dangerous: your process can always, for example, be SIGKILLed. I'm guessing though you're hoping to make sure that, for example, temporary files are cleaned up?

FWIW, this is all very similar to this issue: https://github.com/simonmar/async/issues/41

sol:
Thanks for the pointer to that issue.

PR #273, opened 2016-08-08
Doc tweaks

soenkehahn:
This is using `tinc` in the `Makefile` for building the documentation. Not everyone has `tinc` installed. But my guess is that not many people will actually build the documentation. (I volunteer to try to migrate this to `stack` if wanted.)

@sol: WDYT? I'm trying to get http://hspec.github.io updated.


sol:
This looks great!

Issue #297, opened 2016-12-01
Print QuickCheck labels on success.


andorp:
Usually I use quickcheck properties like

````haskell
it "works for all partition" . forAll arbitrary $ \x ->
    let y = neg x
    in counterexample (x, y) $ case x of
       x | x > 0   -> label "positive"   $ y < 0
       x | x == 0  -> label "zero"       $ y == 0
       x | x < 0   -> label "negative"   $ y > 0
````

Is there a way to print the label distribution for the test in the test results?

andorp:
I have implemented the print labels functionality for Success only.

sol:
On `master`.

PR #298, opened 2016-12-02
Print QuickCheck labels on successful test run.

andorp:
* New field is introduced to the Result Success constructor.
   It is an optional field and could contain any textual information
   for the successful test result.
* The field has one use at the moment. For QuickCheck test cases:
   if a test case computes labels, labels and their occurrence
   information is rendered.

The motivation behind this change: The QuickCheck's property based
testing can be used well with traditional test design. Labels can
be used to mark the test partitions. Observing the distribution
of partitions could be useful when one investigates false
positive test runs.

Fixes: https://github.com/hspec/hspec/issues/297

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Example.hs
+++ b/hspec-core/src/Test/Hspec/Core/Example.hs
@@ -27,7 +27,7 @@ import           Data.Typeable (Typeable)
 import qualified Test.QuickCheck as QC
 import           Test.Hspec.Expectations (Expectation)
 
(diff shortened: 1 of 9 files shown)


sol:
@andorp sorry for the late response.  I rebased this, applied some refactoring (see https://github.com/hspec/hspec/commit/51b7933cf92934f292900a2a6c03cf8c9627d765) and pushed it to `master`.

Thanks a lot!

PR #299, opened 2016-12-09
Inline untangle and mapAround

DaveCTurner:
In trying to track down an unrelated problem, I inlined these two functions as they're quite cryptically point-free and seem to be clearer like this.


sol:
I agree that the code is cryptic.  The good thing is that because of parametricity there is only one possible implementation for these functions.  So we don't really care how the code looks like and rely on the type checker to guarantee correctness.

PR #302, opened 2017-01-14
README.markdown: Add a travis build badge

erikd:
Its nice to have a travis build badge on the readme, but if I follow that link, the build is actually mostly broken. Not sure why.

Update: This PR built without a problem. Not sure what was happening before.


sol:
Thanks a lot!

Issue #301, opened 2017-01-11
Test count output bleeds through to hspec output


erikd:
For one of my hspec test suites I've set `configQuickCheckMaxSuccess = Just 100000`. When the tests are run it now prints `xxxxx/100000` for each test, but if the description for the `prop` test is too short, part of the count gets left behind when the test is completed so I get for example:

```
  complement
  negate0000
  signum0000
  logical and/or/xor
```
where the `0000` after "negate" and "signum" is the last four digits of `100000`.

sol:
@erikd I guess we need to pad with spaces if the description is shorter than the QC count output.  Patches welcome!

erikd:
I've got  a trivial test program that displays the problem:
```
{-# LANGUAGE ScopedTypeVariables #-}
import Control.Monad

import Test.Hspec
import Test.Hspec.QuickCheck
import Test.Hspec.Runner

import System.Exit


main :: IO ()
main = do
  summary <- hspecWithResult config testAll
  when (summaryFailures summary == 0)
    exitSuccess
  exitFailure
  where
    config = defaultConfig { configQuickCheckMaxSuccess = Just 100000 }

testAll :: Spec
testAll =
  prop "short" $ \ (a, b :: Int) ->
    (a + b) `shouldBe` (b + a)
```
but can't figure out where the property description is being printed. Clues?

erikd:
Never mind. The use of git submodules made this a little difficult to figure out but I have a fix now.

erikd:
I've just spent another hour on this problem. I've now come up with two solutions that fix the actual problem I see when running tests, but fail in the test suite.

PR #303, opened 2017-01-14
Add trailing white space to property name output

erikd:
If the test name for a property test was too short, the test count would
bleed through to the actual test output. Fix it by adding a fixed amount
of trailing whitespace.

Closes: https://github.com/hspec/hspec/issues/301

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Formatters.hs
+++ b/hspec-core/src/Test/Hspec/Core/Formatters.hs
@@ -141,7 +141,7 @@ specdoc = silent {
 } where
     indentationFor nesting = replicate (length nesting * 2) ' '
     formatProgress (current, total)
-      | total == 0 = show current ++ "\r"
+      | total == 0 = "                                    \r"
       | otherwise  = show current ++ "/" ++ show total ++ "\r"
 
 


erikd:
Just noticed that the tests are failing. Let me fix that.

sol:
As I understand it this will add unintended line breaks if the description + whitespace padding is longer than the terminal width, at least with some terminals.

I would hope there is a more robust way to solve this.

erikd:
The problem is that finding the place to fix this was not that easy (mainly due to unfamiliarity with the code), but I've got no idea how to plumb the total count through to the place that needs it.

sol:
Instead of plumbing it through, could we erase it where it is written in the first place?

erikd:
I think that may have fixed it. Lets see what travis says.

hspec-core/src/Test/Hspec/Core/Formatters.hs
```diff
@@ -141,7 +141,7 @@ specdoc = silent {
 } where
     indentationFor nesting = replicate (length nesting * 2) ' '
     formatProgress (current, total)
-      | total == 0 = show current ++ "\r"
+      | total == 0 = "                                    \r"
```
sol: I can't imagine that this solves the issue.

erikd: Yes, you're right. I forgot to add `-fforce-recomp` to the command line.

erikd:
This code is extremely frustrating to work on. Every time I come back to this I make the mistake of modifying code in `hspec-core` only to find out some 10-20 minutes later that my changes are not being picked up because I installed a previous version of that package in the sandbox. The project needs "how to hack on this thing" documentation, so people like me don't waste their time chasing their own tail.

I don't have time to work on this at the moment. I would still like to see this fixed sooner rather than later.

PR #318, opened 2017-06-05
Make SpecM and FormatM an instance of MonadCatch, MonadThrow.

mboes:


sol:
@mboes Thank you for taking the time to open this PR!

`FormatM` will be replaced with a new API for formatters (see https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Core/Format.hs).  This new API allows you to use your own monad stack for your formatter, and of course everybody is free to add `MonadCatch` and `MonadThrow` instances for their own monad stacks.

With this in mind, this change would only be relevant for `SpecM`.

@mboes Before making a final call on this, may I ask you for your usage example for the `MonadCatch`/`MonadThrow` instances for `SpecM`?

sol:
@mboes closing as "wontfix" for now.  Please still feel free to answer my question and I may reopen.

mboes:
Sorry, didn't see your question earlier. I actually only care about `SpecM`. I updated `FormatM` as well for consistency, but that's not needed for my use case.

The use case is inline-java. I write tests that use combinators `push` and `pop` both of which want a `MonadIO` that for various reasons (they need to cleanup stuff if an exception happens) should also be a `MonadCatch`/`MonadThrow`. Currently, I can't write tests involving those combinators because `SpecM` is only `MonadIO`.

sol:
@mboes Sorry for the late reply. I still don't understand for what reason your code would live in `SpecM`. Can you motivate your use case with a minimal example here?

mboes:
I have a combinator with the following type:

```haskell
push :: (MonadCatch m, MonadIO m) => m (Pop a) -> m a
```

I want to test properties of this combinator. Some of these properties are about the lifetimes of objects created by the inner action provided as an argument. So if I want to assert anything of interest, I need to do so from inside the inner action, which in turn forces the instantiation of `m` above to `SpecM`. I could work around this by passing values out of the inner action and writing assertions on that. But that feels like an inconvenient limitation to me.

sol:
@mboes Thanks for elaborating.  I'm very much sorry, I still don't understand what this is about.  Can you please provide a minimal example (actual source code) that illustrates your use case.

sol:
@mboes to further elaborate, actual test code in Hspec is supposed to live in `IO`.  `SpecM` is a monad to construct a test tree.

Closing for inactivity.

PR #326, opened 2017-08-27
doc: Improve monospace font stack on macOS

ear:
Hello. macOS does not come with `Liberation Mono` installed thus the current documentation monospace font stack falls back to `Courier New`. As you can see in the screenshot below it looks _extremely_ thin. I thought I'd try to add a couple better options using only fonts installed by default on this platform.

Courier New
---

This is the present:

screen shot 2017-08-27 at 17 20 55

Menlo
---

It's the default for the Apple's shell program `Terminal.app` it has been in OS X forever. Incidentally https://play.golang.org uses it too.

screen shot 2017-08-27 at 17 20 59

Courier
---

If `Menlo` is not available at the very least try `Courier` before `Courier New`. It looks like this:

screen shot 2017-08-27 at 17 21 02

Thanks for taking the time to look through this.

### DIFF
--- a/doc/css/custom.less
+++ b/doc/css/custom.less
@@ -70,7 +70,7 @@ h2 {
 
 pre, code, kbd, samp, tt {
   // font-family: Cousine, "Liberation Mono", "Courier New", monospace;
-  font-family: "Liberation Mono", "Courier New", monospace;
+  font-family: "Liberation Mono", Menlo, Monaco, "Courier New", monospace;
   .font-size(12);
   background-color: #F8F8F8;
 }


sol:
@ear Hey! This looks indeed much nicer! Just to make sure, above you talking about "Courier" as a second fallback, however, the PR specifies "Monaco" as the second fallback.  Is this by intention?

ear:
Oh yes @sol sorry for the confusion! My first attempt had `Courier` between `Monaco` and `Courier New`. Then I realized that it would make it look bad on non-mac systems, because the `New` version looks much better there.

… but I forgot to change the description of the `Courier` paragraph to reflect that realization.

Thanks for merging it so fast!

sol:
Cool! Thanks for your contribution!

PR #327, opened 2017-09-06
Add docs on how to configure the quickcheck driver

futtetennista:
It took me quite a bit of digging to find out it was possible but I was glad to find out that it's possible to modify the default configuration 🙂


sol:
Thanks a lot!

PR #337, opened 2017-12-20
Change hspec-discover sort from lexicographic to to natural order

ephemient:
I would like my tests to be [sorted for humans](https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/). It seems easy to add to `hspec-discover`, would you consider something like this?

### DIFF
--- a/hspec-discover/src/Test/Hspec/Discover/Sort.hs
+++ b/hspec-discover/src/Test/Hspec/Discover/Sort.hs
@@ -3,36 +3,20 @@ module Test.Hspec.Discover.Sort (
 , naturalSortKey
 ) where
 
+import           Control.Arrow
 import           Data.Char
 
 data NaturalSortChunk
-  = NumericChunk { chunkNum :: Integer, chunkStr :: String }
-  | StringChunk { chunkStr :: String }
-  deriving (Eq, Show)
+  = NumericChunk { chunkNum :: Integer, chunkStrLen :: Int }
+  | StringChunk { chunkLCStr :: String }
+  deriving (Eq, Ord, Show)
 
-instance Ord NaturalSortChunk where
-  compare (NumericChunk n1 s1) (NumericChunk n2 s2) =
-    compare n1 n2 ~> compareLength s1 s2 ~> compare s1 s2
-  compare (StringChunk s1) (StringChunk s2) = compare s1 s2
-  compare (NumericChunk _ _) (StringChunk _) = LT
-  compare (StringChunk _) (NumericChunk _ _) = GT
-
-(~>) :: Ordering -> Ordering -> Ordering
-EQ ~> a = a
-a ~> _ = a
-infixr 5 ~>
-
-compareLength :: [a] -> [b] -> Ordering
-compareLength [] [] = EQ
-compareLength [] _ = LT
-compareLength _ [] = GT
-compareLength (_:xs) (_:ys) = compareLength xs ys
-
-naturalSortKey :: String -> [NaturalSortChunk]
-naturalSortKey s@(c:_)
-  | isDigit c = NumericChunk (read num) num : naturalSortKey afterNum
-  | otherwise = StringChunk             str : naturalSortKey afterStr
-  where
-    (num, afterNum) = span  isDigit s
-    (str, afterStr) = break isDigit  s
-naturalSortKey _ = []
+naturalSortKey :: String -> ([NaturalSortChunk], String)
+naturalSortKey = chunks &&& id where
+  chunks s@(c:_)
+    | isDigit c = NumericChunk (read num) (length num) : chunks afterNum
+    | otherwise = StringChunk (map toLower str) : chunks afterStr
+    where
+      (num, afterNum) = span  isDigit s
+      (str, afterStr) = break isDigit s
+  chunks _ = []
--- a/hspec-discover/src/Test/Hspec/Discover/Sort.hs
+++ b/hspec-discover/src/Test/Hspec/Discover/Sort.hs
@@ -1,0 +1,38 @@
+module Test.Hspec.Discover.Sort (
+  NaturalSortChunk(..)
+, naturalSortKey
+) where
+
+import           Data.Char
+
+data NaturalSortChunk
+  = NumericChunk { chunkNum :: Integer, chunkStr :: String }
+  | StringChunk { chunkStr :: String }
+  deriving (Eq, Show)
+
+instance Ord NaturalSortChunk where
+  compare (NumericChunk n1 s1) (NumericChunk n2 s2) =
+    compare n1 n2 ~> compareLength s1 s2 ~> compare s1 s2
+  compare (StringChunk s1) (StringChunk s2) = compare s1 s2
+  compare (NumericChunk _ _) (StringChunk _) = LT
+  compare (StringChunk _) (NumericChunk _ _) = GT
+
+(~>) :: Ordering -> Ordering -> Ordering
+EQ ~> a = a
+a ~> _ = a
+infixr 5 ~>
+
+compareLength :: [a] -> [b] -> Ordering
+compareLength [] [] = EQ
+compareLength [] _ = LT
+compareLength _ [] = GT
+compareLength (_:xs) (_:ys) = compareLength xs ys
+
+naturalSortKey :: String -> [NaturalSortChunk]
+naturalSortKey s@(c:_)
+  | isDigit c = NumericChunk (read num) num : naturalSortKey afterNum
+  | otherwise = StringChunk             str : naturalSortKey afterStr
+  where
+    (num, afterNum) = span  isDigit s
+    (str, afterStr) = break isDigit  s
+naturalSortKey _ = []
--- a/hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
+++ b/hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
@@ -5,6 +5,7 @@ import           Helper
 import           System.IO
 import           System.Directory
 import           System.FilePath
+import           Data.List (sort)
 
 import           Test.Hspec.Discover.Run hiding (Spec)
 import qualified Test.Hspec.Discover.Run
@@ -29,10 +30,6 @@ spec = do
           "{-# LINE 1 \"test-data/nested-spec/Spec.hs\" #-}"
         , "{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}"
         , "module Main where"
-        , "import qualified Dir8.X64Spec"
-        , "import qualified Dir8.X128Spec"
-        , "import qualified Dir16.X48Spec"
-        , "import qualified Dir16.X160Spec"
         , "import qualified Foo.Bar.BazSpec"
         , "import qualified Foo.BarSpec"
         , "import qualified FooSpec"
@@ -41,11 +38,7 @@ spec = do
         , "main = hspec spec"
         , "spec :: Spec"
         , "spec = " ++ unwords [
-               "postProcessSpec \"test-data/nested-spec/Dir8/X64Spec.hs\" (describe \"Dir8.X64\" Dir8.X64Spec.spec)"
-          , ">> postProcessSpec \"test-data/nested-spec/Dir8/X128Spec.hs\" (describe \"Dir8.X128\" Dir8.X128Spec.spec)"
-          , ">> postProcessSpec \"test-data/nested-spec/Dir16/X48Spec.hs\" (describe \"Dir16.X48\" Dir16.X48Spec.spec)"
-          , ">> postProcessSpec \"test-data/nested-spec/Dir16/X160Spec.hs\" (describe \"Dir16.X160\" Dir16.X160Spec.spec)"
-          , ">> postProcessSpec \"test-data/nested-spec/Foo/Bar/BazSpec.hs\" (describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec)"
+               "postProcessSpec \"test-data/nested-spec/Foo/Bar/BazSpec.hs\" (describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec)"
           , ">> postProcessSpec \"test-data/nested-spec/Foo/BarSpec.hs\" (describe \"Foo.Bar\" Foo.BarSpec.spec)"
           , ">> postProcessSpec \"test-data/nested-spec/FooSpec.hs\" (describe \"Foo\" FooSpec.spec)"
           ]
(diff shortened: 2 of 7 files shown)


sol:
@ephemient Hey, thanks for taking the time to open this PR!

> I would like my tests to be sorted for humans. It seems easy to add to hspec-discover, would you consider something like this?

From my perspective we can consider to make this the default behavior.  However, `natural-sort` has transitive dependencies on `text` and `parsec` which I would prefer to prevent.

What we would need is a standalone implementation of natural sorting that we vendor with Hspec.

sol:
One more thing, contributions to Hspec require tests prior to being considered ;)

ephemient:
@sol I've updated the PR to drop the dependency, replacing it with a home-grown natural sort, and added a `Spec` for it. Well, just the sorting function; I haven't dug into how to add it to `RunSpec` or the integration tests yet.

hspec-discover/src/Test/Hspec/Discover/Sort.hs
```diff
@@ -0,0 +1,38 @@
...
+  compare (NumericChunk n1 s1) (NumericChunk n2 s2) =
+    compare n1 n2 ~> compareLength s1 s2 ~> compare s1 s2
+  compare (StringChunk s1) (StringChunk s2) = compare s1 s2
+  compare (NumericChunk _ _) (StringChunk _) = LT
+  compare (StringChunk _) (NumericChunk _ _) = GT
+
+(~>) :: Ordering -> Ordering -> Ordering
+EQ ~> a = a
+a ~> _ = a
+infixr 5 ~>
```
sol: - I think this is the same as `mappend` / `<>`.
- Even if we were to implement this ourself, I don't see the benefit of defining a operator for one-way use.  What I would want instead is a name that conveys what this function does.

hspec-discover/src/Test/Hspec/Discover/Sort.hs
```diff
@@ -0,0 +1,38 @@
...
+
+compareLength :: [a] -> [b] -> Ordering
+compareLength [] [] = EQ
+compareLength [] _ = LT
+compareLength _ [] = GT
+compareLength (_:xs) (_:ys) = compareLength xs ys
```
sol: This can be replaced with

```haskell
compareLength :: [a] -> [a] -> Ordering
compareLength = comparing length
```

hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
```diff
@@ -38,7 +41,11 @@ spec = do
         , "main = hspec spec"
         , "spec :: Spec"
         , "spec = " ++ unwords [
-               "postProcessSpec \"test-data/nested-spec/Foo/Bar/BazSpec.hs\" (describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec)"
+               "postProcessSpec \"test-data/nested-spec/Dir8/X64Spec.hs\" (describe \"Dir8.X64\" Dir8.X64Spec.spec)"
+          , ">> postProcessSpec \"test-data/nested-spec/Dir8/X128Spec.hs\" (describe \"Dir8.X128\" Dir8.X128Spec.spec)"
+          , ">> postProcessSpec \"test-data/nested-spec/Dir16/X48Spec.hs\" (describe \"Dir16.X48\" Dir16.X48Spec.spec)"
+          , ">> postProcessSpec \"test-data/nested-spec/Dir16/X160Spec.hs\" (describe \"Dir16.X160\" Dir16.X160Spec.spec)"
+          , ">> postProcessSpec \"test-data/nested-spec/Foo/Bar/BazSpec.hs\" (describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec)"
```
sol: I think there is no really need to test this here.  Unit tests for the sort function are good enough.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+      it "numbers first" $ do
+        "Hello2World" <=> "Hello World" `shouldBe` LT
+        "Hello World" <=> "Hello2World" `shouldBe` GT
+      it "string segments" $ do
+        "Hello2World9" <=> "Hello2World!0" `shouldBe` LT
+        "Hello2World!0" <=> "Hello2World9" `shouldBe` GT
+      it "numeric segments" $ do
+        "3.1.415" <=> "3.14.15" `shouldBe` LT
+        "3.14.15" <=> "3.1.415" `shouldBe` GT
+      it "numeric ties by string" $ do
```
sol: I think this is the right thing to do :clap:

Many other implementations for natural sorting miss this case and as a result assume that e.g. `00` and `0` are equal.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+        sortBy (<=>)
+          [ "z" ++ show (n :: Int) ++ ".txt"
+          | n <- [1, 10] ++ [100..102] ++ [11..19] ++ [2] ++ [20] ++ [3..9]
+          ] `shouldBe`
+          [ "z" ++ show (n :: Int) ++ ".txt"
+          | n <- [1..20] ++ [100..102]
+          ]
+  where
+    (<=>) = comparing naturalSortKey
+    infix 4 <=>
```
sol: Not too impressed by this spec.  May I ask you if you developed the code test-driven?  Or did the tests come as an afterthought?

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+        "Hello2World" <=> "" `shouldBe` GT
+      it "numbers first" $ do
+        "Hello2World" <=> "Hello World" `shouldBe` LT
+        "Hello World" <=> "Hello2World" `shouldBe` GT
+      it "string segments" $ do
+        "Hello2World9" <=> "Hello2World!0" `shouldBe` LT
+        "Hello2World!0" <=> "Hello2World9" `shouldBe` GT
+      it "numeric segments" $ do
+        "3.1.415" <=> "3.14.15" `shouldBe` LT
+        "3.14.15" <=> "3.1.415" `shouldBe` GT
```
sol: You may be able to make some of the things you test here redundant by deriving the `Ord` instance, e.g.:

```haskell
import           Data.Char

data NaturalSortChunk = NumericChunk Integer Int | StringChunk String
  deriving (Eq, Show, Ord)

naturalSortKey :: String -> [NaturalSortChunk]
naturalSortKey s@(c : _)
  | isDigit c = NumericChunk (read num) (length num) : naturalSortKey afterNum
  | otherwise = StringChunk str : naturalSortKey afterStr
  where
    (num, afterNum) = span  isDigit s
    (str, afterStr) = break isDigit  s
naturalSortKey [] = []
```

hspec-discover/src/Test/Hspec/Discover/Sort.hs
```diff
@@ -0,0 +1,38 @@
...
+
+naturalSortKey :: String -> [NaturalSortChunk]
+naturalSortKey s@(c:_)
+  | isDigit c = NumericChunk (read num) num : naturalSortKey afterNum
+  | otherwise = StringChunk             str : naturalSortKey afterStr
```
sol: Aligning code, while it can look nice, has it's downsides.  Here, if you change the `isDigit c` case, then you may also need to change the `otherwise` case to keep things aligned, which gives you a bigger diff, increases the chance for merge conflicts  and puts more burden on the code reviewer.  This is why I try to avoid aligning out of pragmatism.

sol:
If we already tackle this, we may also want to sort alphabetically, e.g `A a B b` instead of `A B a b`.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,49 @@
...
+
+spec :: Spec
+spec = do
+  describe "naturalSortKey" $ do
+    context "compares" $ do
+      it "reflexively" $ do
+        "" <=> "" `shouldBe` EQ
+        "Hello2World" <=> "Hello2World" `shouldBe` EQ
+        "123Hello456" <=> "123Hello456" `shouldBe` EQ
+        "" <=> "" `shouldBe` EQ
```
sol: Provided you derive the instance, I don't think you will need this test.  But if it were that you really want to test this, then this would be a prime use case for QuickCheck, e.g.:

```haskell
it "is reflexive" $ do
  property $ \ x -> x <=> x `shouldBe` EQ
```

hspec-discover/src/Test/Hspec/Discover/Sort.hs
ephemient: Oh I didn't realize there was a Monoid instance for Ordering. That's exactly what this is, great.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
ephemient: Ah sure, I suppose since `isDigit` is ASCII digits only it's not like there can be ties that need to be broken by string.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
ephemient: It was only after writing this test that I realized I wanted to add it to the comparator :)

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
ephemient: Satisfying the `z(\d+).txt` image in the blog post I linked to was goal as I was writing this, although it was tested in GHCi without this spec at first, then I added the ones above as they were also things I was checking interactively.

ephemient:
I've been distracted by [AoC](http://adventofcode.com/) and the holidays, but I'll be back to continue cleaning up this PR. Thanks for looking.

> If we already tackle this, we may also want to sort alphabetically, e.g A a B b instead of A B a b.

I thought of that, but held off because there's a tarpit of potential locale-based issues that I'd defer to Text on.

On the other hand, since these are module names maybe there's no such issues.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
sol: The code I give still uses the string length as a tie breaker.  This is why is `NumericChunk Integer Int`.  The `Int` is the length of the string.

sol: And yes, there are ties, e.g. `01` and `1`, see below.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
sol: Don't test interactively, use [`hspec/sensei`](https://github.com/hspec/sensei) ;)

sol:
> I thought of that, but held off because there's a tarpit of potential locale-based issues that I'd defer to Text on.

I assume changing to `StringChunk String String` and then doing `StringChunk (map toLower str) str` will do the trick with a derived `Ord` instance.

sol:
> I've been distracted by AoC and the holidays

No worries, no need to rush ;)

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
ephemient: I meant, broken by string comparison in addition to by length. That could happen with Unicode digits but (numeric value, length) should be sufficient with ASCII only.

ephemient:
> I assume changing to `StringChunk String String` and then doing `StringChunk (map toLower str) str` will do the trick with a derived `Ord` instance.

Not quite the behavior I would like - I changed it to postpone the original string comparison until after all the chunks are compared.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,57 @@
...
+spec :: Spec
+spec = do
+  describe "naturalSortKey" $ do
+    context "compares" $ do
+      it "reflexively" $ property $ \a -> do
+        a <=> a `shouldBe` EQ
+      it "symmetrically" $ property $ \a b -> do
+        a <=> b === compare EQ (b <=> a)
+      it "transitively" $ property $ \a b c -> do
+        a <=> b == b <=> c ==> a <=> b === a <=> c
```
sol: I think eflexivity, symmetry and transitivity is always given for derived instances, so there is no need to test for this.

We might want to test for injectivity though, see https://github.com/hspec/hspec/commit/62e0c4437c7fb88179491941081257b9cdbb8096.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
```diff
@@ -0,0 +1,57 @@
...
+        "3.14.15" <=> "3.1.415" `shouldBe` GT
+      it "numeric ties by string length" $ do
+        "Hello 02 World" <=> "Hello 002 World" `shouldBe` LT
+        "Hello 002 World" <=> "Hello 02 World" `shouldBe` GT
+      it "case squashing" $ do
+        "Helloworld1" <=> "HelloWorld2" `shouldBe` LT
+        "HelloWorld2" <=> "helloworld1" `shouldBe` GT
+      it "string ties by case" $ do
+        "HelloWorld3" <=> "Helloworld3" `shouldBe` LT
+        "Helloworld3" <=> "HelloWorld3" `shouldBe` GT
```
sol: https://github.com/hspec/hspec/commit/11837f69e345092897d3b360164f6b2f07747225

I think for derived `Ord` instances `a < b <=> b > a` is always true.

hspec-discover/test/Test/Hspec/Discover/SortSpec.hs
ephemient: Good point. Thanks for merging and cleaning up 😃

sol:
@ephemient looking at the following example:

```
SpecFoo.hs
SpecFoo.lhs
Specfoo.hs
Specfoo.lhs
```

we want that order.  We don't want the spec ordering depend on the file extension.   But the current implementation would give us:

```
SpecFoo.hs
Specfoo.hs
SpecFoo.lhs
Specfoo.lhs
```

I changed this, see https://github.com/hspec/hspec/commit/65b246ef0840e5b3f4cf3567c741d1edc33debb8.

ephemient:
Ah hmm, so the reason I did it differently is because I didn't like `FooBar2 < Foobar1`. But I agree that having modules sorted by extension is worse.
I suppose another workaround could be to compare without the extension first. But that seems more complex than necessary.

Issue #336, opened 2017-11-29
Missing Semigroup instance on GHC HEAD


harpocrates:
As part of the semigroup-monoid proposal, `Semigroup` becomes a superclass of `Monoid`. This is due to happen in 8.4.

It looks like `hspec` still has a `Monoid` instance [here](https://github.com/hspec/hspec/blob/51b7933cf92934f292900a2a6c03cf8c9627d765/hspec-core/src/Test/Hspec/Core/Runner.hs#L239) without a `Semigroup` instance, so `hspec` doesn't compile on GHC HEAD.

sol:
Hi @harpocrates , thank you for taking the time to open this issue.  A PR would be very much welcome.  I can push a minor release to Hackage once this is fixed.

DanBurton:
Bumped into this trying to build with the ghc alpha.

```
    [22 of 22] Compiling Test.Hspec.Core.Runner ( src/Test/Hspec/Core/Runner.hs, .stack-work/dist/x86_64-osx/Cabal-2.1.0.0/build/Test/Hspec/Core/Runner.o )

    /private/var/folders/bz/c3q0fbt94dbgx69km1r5rbxw0000gn/T/stack7703/hspec-core-2.4.4/src/Test/Hspec/Core/Runner.hs:229:10: error:
        • No instance for (Semigroup Summary)
            arising from the superclasses of an instance declaration
        • In the instance declaration for ‘Monoid Summary’
        |
    229 | instance Monoid Summary where
        |          ^^^^^^^^^^^^^^
```

I'll try and whip up a quick little PR.

DanBurton:
I've opened #342 which addresses this.

PR #342, opened 2018-01-05
Add Semigroup instance for Summary, addresses hspec/hspec#336

DanBurton:
Pretty straightforward migration, as suggested by:
- https://ghc.haskell.org/trac/ghc/wiki/Migration/8.4#SemigroupMonoidsuperclasses
- https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#Writingcompatiblecode
  
[edit]
This change conditionally imports Data.Semigroup from base, if the base version is new enough (>= 4.9) to have it, and provides an instance for it. If the base version is older (< 4.9), then the code is equivalent to what it was before, providing no Semigroup instance. This is one of the suggested migration patterns put forward by the migration guide for those who do not wish to incur a dependency on `semigroups`.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -27,6 +27,7 @@ import           Test.Hspec.Core.Compat
 
 import           Control.Monad
 import           Data.Maybe
+import qualified Data.Semigroup as Sem
 import           System.IO
 import           System.Environment (getProgName, getArgs, withArgs)
 import           System.Exit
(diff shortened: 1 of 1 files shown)


DanBurton:
Well the good news is it passed for GHC >= 8.0.2. I'll work on some CPP (should be easy, as it is already suggested in the links above) for compat with older versions.

sol:
@DanBurton Hey, thanks for looking into this.  My main issues with this change are:

- Nested `CPP`
- Not easily testable with the latest version of GHC due to other packages not compiling yet

Both is not your fault.  But from what I checked, the `Monoid` instance is unused.  If this is indeed the case we may prefer to remove the instance instead.

sol:
@DanBurton Thanks for your contribution.  I chose a more lightweight solution for this.

I'm still tempted to remove the `Monoid` instance.  The reason I did it like this for now is that this can be released as a minor release.

sol:
Released to Hackage as `2.4.6`.

DanBurton:
No complaints here! Cheers.

PR #348, opened 2018-03-09
hspec-discover.md: make example compile

domenkozar:


domenkozar:
Ah, the problem was that cabal main-is should be `Main.hs` not `Spec.hs`.

PR #350, opened 2018-03-12
Evaluates examples Unmasked (close #349)

### LINKED ISSUE #349: Tests are run with masking state set to MaskedInterrutible
The following test program demonstrates that `hspec` runs tests with the masking state set to `MaskedInterruptible` even when exceptions are unmasked before running the tests.

```
import Control.Exception
import Control.Monad.IO.Class
import Test.Hspec

main :: IO ()
main = do
  print =<< getMaskingState
  hspec $ do
    describe "foobar" $ do
      it "foobar" $ do
        print =<< getMaskingState
```

I am not sure if this is intentional or not but it is at least a bit unexpected so if it is intentional it would be nice to document this.

I tested this using `hspec-2.4.8`.

sol:


hspec-core/src/Test/Hspec/Core/Util.hs
```diff
@@ -129,13 +129,4 @@ formatException err@(SomeException e) = case fromException err of
...
-  where
-    runAction = async ((action >>= evaluate))
-    waitForAction = waitCatch
-    cancelAction a = do
-      cancel a
-      -- It is important to wait here to make sure all finalizers in action have
-      -- been run. Otherwise the main thread can exit before they have finished
-      -- and the finalizers are only partially run.
-      waitCatch a -- We use waitCatch to hide the ThreadKilled exception
+safeTry action = withAsync (action >>= evaluate) waitCatch
```
sol: @cocreature FYI, this will still retain your desired behavior.  `async` takes care of this by now, and it does so while getting masking right at the same time.

PR #354, opened 2018-05-28
Add Profunctor instance for SpecM

### LINKED ISSUE #353: SpecM should be a Profunctor
As far as I understand it, the `SpecM a r` values are tests that take `a` as input and return `r`. This makes them lawful instances of [`Profunctor`](https://www.stackage.org/haddock/lts-11.11/profunctors-5.2.2/Data-Profunctor.html).

A use case I have for that is running a test that requires input with a single `before` but different actual arguments:

```haskell
-- | Perform some common checks on a list
listChecks = do
  it "is non empty" $ \lst -> not (null lst)
  it "is not too long" $ \lst -> length lst < 10

-- | Something to verify the results of
expensiveComputation :: ([a], [b])

-- | All the tests
spec = do
  beforeAll expensiveComputation $
    context "first list" $ lmap fst listChecks
    context "second list" $ lmap snd listChecks
```

koterpillar:
Fixes #353.

I think with these changes some of the hooks like `around` might have a better implementation, and I'll explore this further.

### DIFF
--- a/hspec-core/hspec-core.cabal
+++ b/hspec-core/hspec-core.cabal
@@ -2,7 +2,7 @@
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: cd0a31bd7edaf1f1a446581fe294f3b1ece7f6ad4c2b328ed847b570055241fa
+-- hash: b81b94506310bc08d6e1282c17048835790a2636944d63f65ea326efcc123a34
 
 name:             hspec-core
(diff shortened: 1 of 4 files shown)


sol:
> I think with these changes some of the hooks like around might have a better implementation, and I'll explore this further.

I would be very much interested in any simplifications ;)

PR #355, opened 2018-06-05
Add build-tool-depends to hspec-discover:spec

Ericson2314:
See http://cabal.readthedocs.io/en/latest/developing-packages.html?highlight=build-tools#pkg-field-build-tool-depends. This is proper way to make the test suite ensure that executable is on the PATH.

### DIFF
--- a/hspec-discover/package.yaml
+++ b/hspec-discover/package.yaml
@@ -47,3 +47,6 @@ tests:
       - hspec-discover
       - hspec-meta >= 2.3.2
       - QuickCheck >= 2.7
+    verbatim: |
+      build-tool-depends:
+          hspec-meta:hspec-meta-discover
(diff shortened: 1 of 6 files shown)


Ericson2314:
This works now.

hspec-discover/package.yaml
```diff
@@ -47,3 +47,6 @@ tests:
       - hspec-discover
       - hspec-meta >= 2.3.2
       - QuickCheck >= 2.7
+    verbatim: |
+      build-tool-depends:
+          hspec-meta:hspec-meta-discover
```
sol: It is more robust to use:

```
verbatim:
  build-tool-depends: ...
```
Specifically use a mapping instead of a string for `verbatim`.  A string will just get included literally in the generated `.cabal` file, on the other hand with a mapping `build-tool-depends` overrides any existing fields with the same name.  This is important as `hpack-0.30.0` will add `build-tool-depends` in certain situations.

sol:
@Ericson2314 thanks taking the time to open this issue.  I'm closing this in favor of https://github.com/hspec/hspec/pull/368.

PR #360, opened 2018-06-29
Add mapSubject and ignoreSubject

sol:


koterpillar:
Like! I was going to do a similar thing after your point about including so many dependencies.

PR #369, opened 2018-08-04
Fix you -> your typo in doc/options.md

tombusby:
"depends on how your run your tests" should be "depends on how you run your tests"


sol:
Nice catch! Thanks a lot!

PR #373, opened 2018-09-03
QuickCheck 2.12 compatibility.

nick8325:
* Remove formatLabels because QC no longer defines formatLabel. (This function wasn't called anyway, it seems.)
* Remove InsufficientCoverage, which no longer exists.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/QuickCheckUtil.hs
+++ b/hspec-core/src/Test/Hspec/Core/QuickCheckUtil.hs
@@ -20,11 +20,6 @@ import           Test.QuickCheck.Random
 
 import           Test.Hspec.Core.Util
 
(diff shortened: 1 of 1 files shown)


sol:
@nick8325 Thanks a lot for this PR.  A couple of more changes are required to make the tests pass.  I'm looking into it.  I'm also adding an upper bound for QuickCheck.

In the future, can you ping me on new QuickCheck releases?  If possible, a heads up even before the release happens would be nice ;).

sol:
@nick8325 just to be 100% clear, I was talking about major releases here ;).

nick8325:
Oh yes, I forgot to run `cabal test`! Sorry about that!

I meant to ping you but forgot in the excitement of making a release :) I'll try to remember in future.

PR #377, opened 2018-09-17
Use `shouldBe` instead of == in property test docs.

seagreen:
If a test fails, this will show "expected ... but got ..." instead of just the value generated by QuickCheck, which is 1000x more useful.

### DIFF
--- a/doc/quickcheck.md
+++ b/doc/quickcheck.md
@@ -11,7 +11,7 @@ turn anything that is a member of the {{'Testable'|id}} class into a
 ```hspec
 describe "read" $ do
   it "is inverse to show" $ property $
-    \x -> (read . show) x == (x :: Int)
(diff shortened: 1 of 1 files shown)


sol:
@seagreen Hey there :wave:

This looks great! Thanks a lot and sorry for the delay!

I rebased your changes (https://github.com/hspec/hspec/pull/467) and I'll merge them as soon as the build is green.

BTW, your email address does not seem to be associated with GH.  If you associated your email with GH your contributions will show up under your name.

https://github.com/settings/emails

seagreen:
> This looks great! Thanks a lot and sorry for the delay!

No worries, there are more important things to do in life than GitHub.

The reason for the email shenanigans is that I've changed addresses since 2018, this commit was made with my old email, but GitHub is now set to recognize my new one.

Also: thank you for the great library!

Issue #315, opened 2017-05-18
include test duration for each test/test section for the standard formatter (feature request)


epsilonhalbe:
Right now the standardoutput of a test case looks like this (this is from a csv parser I use to verify that the testfiles exportet from sap are what the defintitions say they are).
```
SAP.CL001
  FILE A
    succeeds on first line of A.CSV
    succeeds on small file (100 lines): A.CSV
  FILE B
    succeeds on first line of B.CSV
    succeeds on B.CSV
```
It would be nice to have something like
```
SAP.CL001 (...s)
  FILE A (...s)
    succeeds on first line of A.CSV (0.000155541s)
    succeeds on small file (100 lines): A.CSV (0.008695084s)
  FILE B (...s)
    succeeds on first line of B.CSV (0.000205147s)
    succeeds on B.CSV (0.105569871s)
```
to get the duration of each test; I can do a `around_` block for each test that `print`s a time difference (that is actually what I am doing right now) but then I cannot persist the test duration with `-o outfile` without modifying code.

Also I think that if the duration of each test stored, one can notify performance regressions easily with `--failure-report report`. maybe even adding an option `--fail-on-performance 10` which makes a test fail if performance has gone up by more than 10%.

epsilonhalbe:
here is the snippet I use for printing the time in the `around_` block
```
timeIt :: IO () -> IO ()
timeIt act = bracket getCurrentTime
                     (\start -> do end <- getCurrentTime
                                   printf "\ttest-duration: %s\n" (show $ diffUTCTime end start)
                     )
                     (const act)
```

sol:
I'm looking into this.

epsilonhalbe:
Thanks if i can help somehow, let me know

thomasjm:
Hello -- it looks like the support for this is landed in the master branch, but it doesn't seem to be exposed in the latest version on Hackage: https://hackage.haskell.org/package/hspec-core-2.4.8/docs/Test-Hspec-Core-Formatters.html. 

Is the new formatter API you mentioned going to be available soon? Thanks!

sol:
> Is the new formatter API you mentioned going to be available soon?

No ETA yet, I'm very sorry for that.

eborden:
I was just looking for a feature like this. I'd love to be able to produce a report showing:

1. The duration of each `describe`.
  a. Duration spent in hooks.
  b. Duration spent in `it`.
2. The duration of each `it`.
  a. duration spent in `it`.
[...]

PR #385, opened 2018-11-03
Better locations for --print-slow-items

sol:


sol:
https://github.com/hspec/hspec/issues/315

Issue #331, opened 2017-10-03
Investigate diff performance issue with huge strings with a big Levenshtein distance


sol:

sol:
This is currently not actionable, closing.

PR #394, opened 2018-11-20
Improve diff performance

cuklev:
Hi,
I had a test in a project which compared something really big to something small. It made hspec freeze while reporting diff and easily allocated 50GB+.
I think this is related to #331 

I reimplemented the `lcs` function and now the whole thing behaves a lot better.

### DIFF
--- a/hspec-core/hspec-core.cabal
+++ b/hspec-core/hspec-core.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: fef917dc0cb43c6da4719fb2d080d28e01d8bc888b679a3fa1c61559f02d9848
+-- hash: ba41a22bf03e86b2dcfbc3ee96107cb1fde29bbbcd25d4c00e50c30f748b2a4d
 
 name:             hspec-core
 version:          2.6.0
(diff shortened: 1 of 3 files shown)


sol:
@cuklev Hey, this is great!  I see around 30% speedup (in GHCi).  Is this inline with the performance improvements you observed?

I haven't looked at memory usage.

For sure this is an improvement, so that's great!  I think we are still much slower than other implementations (e.g. GNU diff).  I'm puzzled whether we are even in the right big-O class.

cuklev:
It seems that I am using a slightly different version of the algorithm and my change is not as good as I initially thought.

What are you testing with?
```haskell
[1..10000] `shouldBe` []  -- improves from 40s to 0.3s
[1..10000] `shouldBe` [10000,9999..1]  -- gets worse, from 78s to _I don't have enough RAM_
[1..2000]  `shouldBe` [2000,1999..1]  -- gets worse, from 2.4s to 4.1s
```

Also, do you mean https://github.com/seereason/Diff? Why doesn't `hspec` depend on `Diff`?

sol:
As it's not clear to me whether this is a clear improvement I always hesitated to merge it.

Closing.

@cuklev if you can make a case for it, please do so.

PR #397, opened 2018-12-27
Expose runSpec (needed if you want to ignore the cli).

fisx:
I tried with `withArgs []`, but I also have an environment constructor (run by `beforeAll`) that needs the command line intact.


fisx:
#398 is better.

PR #398, opened 2018-12-28
Add runSpec (close #216)

### LINKED ISSUE #216: Add primitive that does not do any handling of command-line arguments to Test.Hspec.Core.Runner

sol:


hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -99,32 +116,51 @@ ensureSeed c = case configQuickCheckSeed c of
 -- | Run given spec with custom options.
 -- This is similar to `hspec`, but more flexible.
 hspecWith :: Config -> Spec -> IO ()
-hspecWith conf spec = do
-  r <- hspecWithResult conf spec
-  unless (isSuccess r) exitFailure
+hspecWith config spec = getArgs >>= readConfig config >>= doNotLeakCommandLineArgumentsToExamples . runSpec spec >>= evaluateSummary
```
fisx:
```suggestion
hspecWith config spec = 
          getArgs
      >>= readConfig config
      >>= doNotLeakCommandLineArgumentsToExamples . runSpec spec
      >>= evaluateSummary
```

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -82,10 +91,18 @@ applyDryRun c
...
--- | Run given spec and write a report to `stdout`.
+-- | Run a given spec and write a report to `stdout`.
 -- Exit with `exitFailure` if at least one spec item fails.
+--
+-- /Note/: `hspec` handles command-line options and reads config files.  This
+-- is not always desired.  Use `runSpec` if you need more control over these
+-- aspects.
 hspec :: Spec -> IO ()
-hspec = hspecWith defaultConfig
+hspec spec =
```
fisx: What's wrong with the old implementation?  How was it different?

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -99,32 +116,51 @@ ensureSeed c = case configQuickCheckSeed c of
...
 
+-- | `True` if the given `Summary` indicates that there were no
+-- failures, `False` otherwise.
 isSuccess :: Summary -> Bool
 isSuccess summary = summaryFailures summary == 0
 
+-- | Exit with `exitFailure` if the given `Summary` indicates that there was at
+-- least one failure.
+evaluateSummary :: Summary -> IO ()
+evaluateSummary summary = unless (isSuccess summary) exitFailure
```
fisx: Would `exitOnFailure` be a better name?

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -99,32 +116,51 @@ ensureSeed c = case configQuickCheckSeed c of
...
+
 -- | Run given spec and returns a summary of the test run.
 --
 -- /Note/: `hspecResult` does not exit with `exitFailure` on failing spec
 -- items.  If you need this, you have to check the `Summary` yourself and act
 -- accordingly.
 hspecResult :: Spec -> IO Summary
-hspecResult = hspecWithResult defaultConfig
+hspecResult spec = getArgs >>= readConfig defaultConfig >>= doNotLeakCommandLineArgumentsToExamples . runSpec spec
```
fisx:
```suggestion
hspecResult = hspecWithResult defaultConfig
```

hspec-core/src/Test/Hspec/Core/Runner.hs
sol: I think we can deprecate `hspecWith` and friends down the road and only have `hspec` and `runSpec`.  For that reason, if somebody looks at the implementation of these I want them to see a future proof way to achieve the same.

hspec-core/src/Test/Hspec/Core/Runner.hs
sol: I'm conflicted.  On the one hand I like `exitOnFailure`, on the other hand `evaluateSummary` may be somewhat more future proof (e.g. if we need to add more stuff to it in the future than just exiting on failure).

I'll consider and make a call by tomorrow.

Issue #328, opened 2017-09-19
Golden tests


phadej:
Does hspec architecture supports adding new type of tests, or should they be added to the core?

I'd like to have so called golden tests in `hspec` too (they are in `tasty`: http://hackage.haskell.org/package/tasty-golden). Motivated by https://github.com/haskell/haddock/pull/668

sol:
@phadej Yes, it's possible to extend `hspec` by adding new instances for the `Example` type class.  What currently is not possible, is adding new command-line options.  So if you need new command-line options, we would need to hard-code them in `hspec-core`.

CristhianMotoche:
Hi @sol. I'd like to know you have been working on this. In case you don't, I'd love to start working on it.

sol:
@CristhianMotoche hey, I haven't worked on this. But I've been looking into allowing to add custom command line options. Not released yet, but this might come in handy for this feature.

You could get started by hard coding any options  for now and we then change that once I finished up my code.

CristhianMotoche:
Great! Thanks for letting me know that @sol

CristhianMotoche:
Hi @sol . I'm sorry for taking a little while in respond. I've been thinking in the interface that we want to export to users. I came up with the following idea:

```haskell
data Golden str = 
  Golden {
    output :: str -- ^ Lazy bytestring output
    write :: FilePath -> str -> IO (), -- ^ How to write into the golden file the file
    read :: FilePath -> IO str
  }

hGolden :: GoldenTest -- ^ Parameters for the Golden Test
               -> GoldenResult -- ^ Possible result of the test execution

data GoldenResult =
   
```

CristhianMotoche:
Hi @sol . I'm sorry for taking a little while in respond. I've been thinking in the interface that we want to export to users. I came up with the following idea:

```haskell
data Golden str = 
  Golden {
    output :: str -- ^ Lazy bytestring output
    write :: FilePath -> str -> IO (), -- ^ How to write into the golden file the file
    read :: FilePath -> IO str, -- ^ How to read the file,
    name :: String, -- ^ Test name (make sure it's unique otherwise it could be override)
    directory :: FilePath -- ^ Directory where you write your tests
  }

instance Example Golden where
 type Arg Golden = ()
 evaluateExample e = evaluateExample (\() -> e)

instance Example (arg -> Golden) where
 type Arg (arg -> Golden) = arg
 evaluateExample golden params action progressCallback =
[...]

PR #400, opened 2019-01-16
[WIP] Add Golden tests

CristhianMotoche:
## Description
So far, I've defined the interface described here: https://github.com/hspec/hspec/issues/328#issuecomment-451485762

Example:

```haskell
describe "myFunc" $
  it "generates the right output" $
     let output = show $ myFunc params
       in defaultGolden "myFunc" output
```

* First execution will create:
```
.hspec/myFunc/golden
```
with the content of `show $ myFunc params`

* In a second execution.
  * In case there's any update:
    * `golden` won't be updated
    * `actual` will be created
    * Test will fail (until the `golden` file is updated _and_ `actual` is deleted)

  * In case there aren't updates:
    * `golden` won't be updated
    * `actual` won't be created
    * Test will succeed

## TODO
I'd like to discuss about these points before working on them:
- [ ] Add more helper functions (maybe: `defaultGoldenShow`, `defaultGoldenText`, `defaultGoldenBS`)
- [ ] Add a flag to update `golden` file with `actual` (So that developers won't have to execute `mv .hspec/[test name]/{actual,golden}` by hand)


sol:
Hey! Sorry for getting back late!

> Add more helper functions (maybe: `defaultGoldenShow`, `defaultGoldenText`, `defaultGoldenBS`)

For me I tend to drive software development by need.  So what I do is add features only at the point I actually want to use them (or accept a patch from somebody who has an actual use case for some specific feature).  While this may not be perfect, it at least avoids the kind of question of "Should we add this?".

> Add a flag to update `golden` file with `actual` (So that developers won't have to execute `mv .hspec/[test name]/{actual,golden}` by hand)

I don't have a strong opinion here, as I, myself, currently don't have an urge for golden tests.  I think I mentioned that before, I'm working on making command-line flags extensible by third parties.  So that will allow you to define new flags in a separate package, e.g. `hspec-golden`.  What I can offer is:

- Try to get my work out soon (but no promise on a timeline, sorry)
- Accept a patch that adds flags specific to `hspec-golden` to `hspec-core` for now

What I assumed is that this will go into a separate package, e.g. `hspec-golden`.  I would prefer if somebody else (say you) maintain that package.  But I'm always happy to help with anything.

hspec-core/.gitignore
```diff
@@ -0,0 +1 @@
+.hspec/id/
```
sol: Hspec reads options from `.hspec` and `~/.hspec`.   So this directory name may conflict with this.

hspec-core/src/Test/Hspec/Core/Golden.hs
```diff
@@ -0,0 +1,75 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ExistentialQuantification #-}
```
sol: `ExistentialQuantification` are rarely needed.  There is often a better way to achieve things.

hspec-core/src/Test/Hspec/Core/Golden.hs
```diff
@@ -0,0 +1,75 @@
...
+import           System.Directory ( createDirectoryIfMissing
+                                  , doesFileExist
+                                  )
+
+
+-- | Golden tests parameters
+
+data Golden str =
+  forall. Eq str =>
+    Golden {
```
sol: Specifically, I think you can use the `Eq` constraint to `runGolden` to get rid of `ExistentialQuantification`.

PR #403, opened 2019-02-06
specdyn formatter - displays tests being run

### LINKED ISSUE #402: Display tests being run
As an alternative to reporting slow tests, it would be helpful if specdoc would print the title of the example being run. If parallel execution is enabled, it would be even better to print *all* the tests in execution.

Currently this is not possible using a custom formatter either because there are no "example started" hooks.

pepeiborra:
I grew frustrated with large spec suites and slow tests at work (most of them golden tests), some of them `parallel`. Eventually I realised the reasons for my frustration were:
 * Being unable to get the name of the slow test,
 * Being unable to see progress of `parallel` tests,
 * Not really understanding how `parallel` works across spec nodes.

So I set to write a new specdoc-like formatter that would address these frustrations, which ended up requiring changes both to the `Runner` and `Eval` modules and to the `Formatters` api. The result is a fairly large pull request that you may well not want to merge, but I'm not sure what else to do with it.

Let me know if there is any way that I can make these changes more palatable for inclusion.

## Structure
The bulk of the change is contained in the topmost two commits:
1. Asynchronous formatting - hidden behind a flag, when enabled it will call the formatter hooks for all the tests in execution, not just the earliest one.
2. New formatter "specdyn" - a new asynchronous formatter which leverages ANSI terminal codes to show all the tests in execution and their progress.

 ## Usage
The changes are quite invasive, but should be unnoticeable to most hspec users unless the "specdyn" formatter is explicitly selected with `-f specdyn`. Since it relies heavily on ANSI terminal codes, specdyn is unlikely to work in the Windows cmd terminal. 

Worth mentioning is that specdyn captures all the stdout and stderr of test execution and displays it after the footer. This is needed due to the way the terminal is rewritten, but it can be a bit surprising compared to specdoc. 

## Status
All the existing tests pass, and I have added a few tests for asynchronous formatting, although the coverage is rather thin. No tests for specdyn as of yet.

Fixes #402


sol:
Hi there! I currently don't have the time to look through this in detail but I'm excited to see new approaches to formatting and I very much encourage experimenting.

The formatter API has grown historically and I shied away from cleaning it up in the past to reduce breakage.

What I imagine for the future is to provide a more basic, low-level and flexible API that can accommodate more use cases (e.g. custom state). The code in https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Core/Format.hs are first steps in that direction.

A downside of this approach is that writing new formatters might be more work and there might be more code duplication between formatters. But I hope it will facilitate experimentation and hopefully remove myself as a bottleneck.

Would you be interested in implementing your formatter by using the interface provided by https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Core/Format.hs as a third party package. If yes, let's try to identify the changes required to accommodate your use case.

All that said, your approach sounds exciting. I'm eager to try and use it myself in the future and if it turn out to be popular we can certainly consider to make it a core part of hspec down the road.

sol:
Somewhat related, a limitation of third-party formatters  is still that they need to be added to a project explicitly and will then affect everyone who works on that project.

I wrote a proposal on how to address this a while ago here: https://github.com/sol/haskell-plugins/blob/master/README.md

However, it's not clear if this is feasible: https://github.com/commercialhaskell/stack/issues/3927#issuecomment-413092652

pepeiborra:
> Somewhat related, a limitation of third-party formatters is still that they need to be added to a project explicitly and will then affect everyone who works on that project.
> 
> I wrote a proposal on how to address this a while ago here: https://github.com/sol/haskell-plugins/blob/master/README.md
> 
> However, it's not clear if this is feasible: [commercialhaskell/stack#3927 (comment)](https://github.com/commercialhaskell/stack/issues/3927#issuecomment-413092652)

Why not just:
```
-- A version of 'hspec' that allows registering additional formatters for selection via the cli
hspecWithFormatters :: IsFormatter f => [(String,f)] -> Spec -> IO ()
```

pepeiborra:
> Hi there! I currently don't have the time to look through this in detail but I'm excited to see new approaches to formatting and I very much encourage experimenting.
> 
> The formatter API has grown historically and I shied away from cleaning it up in the past to reduce breakage.
> 
> What I imagine for the future is to provide a more basic, low-level and flexible API that can accommodate more use cases (e.g. custom state). The code in https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Core/Format.hs are first steps in that direction.
>
> A downside of this approach is that writing new formatters might be more work and there might be more code duplication between formatters. But I hope it will facilitate experimentation and hopefully remove myself as a bottleneck.
> 
> Would you be interested in implementing your formatter by using the interface provided by https://github.com/hspec/hspec/blob/master/hspec-core/src/Test/Hspec/Core/Format.hs as a third party package. If yes, let's try to identify the changes required to accommodate your use case.
>
> All that said, your approach sounds exciting. I'm eager to try and use it myself in the future and if it turn out to be popular we can certainly consider to make it a core part of hspec down the road.

Hi Simon, your suggestion of moving specdyn down to the FormatM makes sense to me, and it would much reduce the impact of the changes. At that point it makes sense to package specdyn on its own. I can do that

Prerequisites for that:
1. Support for asynchronous formatting, included in this PR. I will cancel this PR and submit a new one with just those changes.
2. The ability to define formatters on top of `Format`, morally speaking we need an instance `IsFormatter (Format m)` or more likely the other way around, as suggested by the existing `formatterToFormat` function. Should I have a first stab or is that something you would prefer to do yourself ?

sol:
> Should I have a first stab 

Yes, please

PR #408, opened 2019-02-19
uit - pending undefined test

BebeSparkelSparkel:
Allows pending test to be undefined.


.gitignore
```diff
@@ -1 +1,2 @@
 dist/
+*.swp
```
sol, line 2: This is not project specific and should go into your global `.gitignore`.

https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -95,6 +96,11 @@ specify = it
 xit :: (HasCallStack, Example a) => String -> a -> SpecWith (Arg a)
 xit label action = before_ pending_ $ it label action
 
+-- |
+-- Allows pending test to be undefined.
+uit :: HasCallStack => String -> SpecWith (Arg Bool)
+uit label = xit label (undefined :: Bool)
```
sol, line 102: Given that this can already be expressed with
```haskell
it "..." pending
```
I'm not eager to add an other primitive for it.

Moreover, writing down a large number of pending test cases (a situation where I agree this primitive could be useful) is contrary to TDD (in a way, it's almost like doing waterfall).

@BebeSparkelSparkel if you care about having this defined somewhere, you can open a PR for `hspec-contrib`.

sol (commented):
Hey! Thanks for using Hspec and sorry for the late reply.

I'm not eager to add this to `hspec`.  However, you can open a PR for `hspec-contrib`.

hspec-core/src/Test/Hspec/Core/Spec.hs
BebeSparkelSparkel: I disagree that being able to describe what a function should accomplish and then filling in those tests is the hated waterfall. In TDD you do need to know what you are trying to accomplish and this makes it very visible in the tests of what needs to be accomplished next.

BebeSparkelSparkel:
`it "..." pending` will work fine for me thanks.

Issue #415, opened 2019-05-07
diff algorithm is quadratic and hangs basically forever for long outputs


nh2:
We encountered a situation where a test run just hung with 100% CPU at

```
  /path/to/Filehs:123:4: 
  1) My.Module It works expected
       expected: 
```

After some investigation, we found that the string that would have been printed in the following `but got:` output is `114883` many chars long.

In this case, the diff algorithm, runs so long that an output is never printed.

Some quick benchmarking shows a quadratic relationship:

```
chars  seconds
20000  17.5
15000  10
13000   7
10000   4
```

The `diff` algorithm is invoked here:

https://github.com/hspec/hspec/blob/3c748d47733da17dbe996a37e7ad6ed7b27b75fc/hspec-core/src/Test/Hspec/Core/Formatters.hs#L217

And its docs here

https://github.com/hspec/hspec/blob/3c748d47733da17dbe996a37e7ad6ed7b27b75fc/hspec-core/vendor/Data/Algorithm/Diff.hs#L11-L14

mention _O(N*D)_ complexity, [where](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.4.6927)

> two sequences A and B
> O(ND) time [complexity]
> where
> * N is the sum of the lengths of A and B and
> * D is the size of the minimum edit script for A and B

If `N` is the length of the string (e.g. 114883 in my case), and the edit script is _O(N)_ for the worst-case that most chars are missing, we end up with _O(N²)_.

Can you confirm this is the case?

nh2:
If yes, we probably want to automatically turn off diffing for Strings longer than a certain length.

nh2:
What I don't get is that it says that UNIX `diff` also implements this algorithm.

However, I've never noticed `diff` performing badly even on huge input files (where the second file is very short, like in our scenario).

For example,

```
$ head testfile
a
a
a
a

$ wc -l testfile
33554432 testfile

$ cat test2
a
b
c
a

 % time diff -u testfile test2 | wc -l
33554437
diff -u testfile test2  1,70s user 0,45s system 93% cpu 2,309 total
```

If `diff` was equally quadratic, this should have taken forever, but it did not.

Am I missing something?

sol:
This is related to https://github.com/hspec/hspec/issues/331.

[...]

PR #416, opened 2019-05-08
Interrupt diff calculation after two seconds (see #415)

sol:
@nh2 what do you think about thin change?  Do you think 2 seconds is appropriate?  Is it crucial to print a warning, e.g. "diff calculation took longer than 2 seconds and was interrupted; not showing diff".

This seems to work well within ghci, but not sure about compiled test suites.  I see a chance that `timeout` won't work reliably if you hog a CPU, but I haven't tried.

### DIFF
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -1,3 +1,6 @@
+## Next
+  - Interrupt diff calculation if it takes longer than two seconds (see #415)
+
 ## Changes in 2.7.1
   - Add compatibility with QuickCheck 2.13.1 (see #410)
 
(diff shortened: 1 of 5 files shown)


nh2:
@sol Personally I think 2 seconds is too much.

If you have a test suite with 500 examples, and many failing, you'd sit there for quite a while to get the outputs.

I think it'd be more useful to have it be 200 ms or so.

It should definitely have a printed warning so that people can see why their tests are slow.

> but not sure about compiled test suites. I see a chance that `timeout` won't work reliably if you hog a CPU, but I haven't tried.

I think it should work. `timeout` / `hWaitForInput` has some problems with real IO (https://github.com/hspec/hspec/issues/415), but given that the algorithm in use is pure Haskell (https://github.com/hspec/hspec/blob/3c748d47733da17dbe996a37e7ad6ed7b27b75fc/hspec-core/vendor/Data/Algorithm/Diff.hs#L11-L14) GHC should be able to interrupt it without problem.

sol:
I pushed it "as is" for now.  I think it's an improvement already.  I'm happy to iterate over it in the future.

PR #417, opened 2019-06-14
Add a missing hook: beforeAllWith

thomasjm:
There seems to be a missing hook in `Test.Hspec.Core.Hooks` at the moment: a `beforeAll`-style hook that lets you access the value in the `SpecWith`.

Currently there's `beforeAll`, which lets you initialize the `SpecWith`, and `beforeAll_`, which lets you run an `IO ()` action. 

This PR adds `beforeAllWith`, which looks like `beforeWith` but only happens before the first spec item. It uses the same `memoize` approach to make this happen. What do you think?


sol:
@thomasjm thanks a lot for working on Hspec.  @neilmayhew added a more DRY version in #447.  It's on Hackage now.

PR #439, opened 2020-08-24
Update docs - module name should be consistent

ekamil:
Module name passed to hspec-discover should be the same as the one in `Main.hs` file, apparently.

### DIFF
--- a/doc/hspec-discover.md
+++ b/doc/hspec-discover.md
@@ -85,7 +85,7 @@ default formatter:
 
 ```haskell
 -- file test/Spec.hs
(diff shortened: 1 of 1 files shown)


sol:
@ekamil what makes you believe that the current documentation is "wrong" somehow?

ekamil:
In short - the example in docs didn't compile.

Using examples given in the docs lead me to 
```
<no location info>: error:
    output was redirected with -o, but no output will be generated
because there is no Main module.
```
after changing _only_ `--module-name=Main` - tests compiled and were discovered.

sol:
@ekamil I'm not sure where the misunderstanding is.  The short story is that unless you know what you are doing, you don't need to specify `--module-name` at all (`Main` is the default).

PR #443, opened 2020-09-16
fix link

ChefYeum:
Noticed that the link to `hspecWith` is broken in the [doc](http://hackage.haskell.org/package/hspec-core-2.7.4/docs/Test-Hspec-Core-Formatters.html#t:Formatter)


sol:
@ChefYeum nice catch! Thanks a lot!

Issue #304, opened 2017-01-20
shouldFail spectree


NorfairKing:
Is there a way to tell hspec 'somewhere in this spectree, a test should fail'.

I'm writing a library that should help with hspec testing: https://github.com/NorfairKing/validity.
I would like to test my library by showing that tests fail when they should.

Any advice?

sol:
I think currently this functionality is missing.  But I think with the help of `mapSpecItem_` this should be easy to add.  PRs accepted.

PR #445, opened 2020-09-23
add shouldFail

mwotton:
this solves https://github.com/hspec/hspec/issues/304 - hopefully it's along the lines you intended, @sol .

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Spec.hs
+++ b/hspec-core/src/Test/Hspec/Core/Spec.hs
@@ -28,6 +28,7 @@ module Test.Hspec.Core.Spec (
 , parallel
 , sequential
 
+, shouldFail
 -- * The @SpecM@ monad
 , module Test.Hspec.Core.Spec.Monad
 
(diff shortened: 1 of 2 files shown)


sol:
@mwotton contrary to what I said over 3 years ago in #304, I'm puzzled if we should add this to `hspec-contrib` instead (at least for now).  What do you think?

(read: Is this functionality require by enough users so that it warrants increasing the API surface for everybody?)

mwotton:
i think hspec-contrib is more suitable, but that would require exposing mapSpecItem_, no?

mwotton:
really, the only people who need this functionality are writing tests-for-tests, and to be completely honest if mapSpecItem_ was available i probably would have just written it as a private function and forgotten to PR it :)

sol:
@mwotton `mapSpecItem_` is re-exported from `Test.Hspec.Core.Spec`.

https://hackage.haskell.org/package/hspec-core-2.7.4/docs/Test-Hspec-Core-Spec.html#v:mapSpecItem_

Keep it as a private function, or open a PR for `hspec-contrib`, whatever is more convenient for you ;).

mwotton:
oh duh, i must have misread :) cheers, i'll open a PR later.

PR #450, opened 2020-11-04
Update doc, the runtime told me to use -N instead of -j

jappeace:
This didn't work for me:

```
✦ ❯ ./result/bin/somebin-test +RTS -N -j4 -RTS
```

This did:

```
✦ ❯ ./result/bin/somebin-test +RTS -N4 -RTS
```

Perhaps I'm wrong but I think the -j option is gone now?


sol:
@jappeace `-j` is an Hspec option (as opposed to an RTS option).  #453 tries to make this more clear.

PR #492, opened 2021-04-16
Add getFinalCount to Formatter API

rhendric:
getFinalCount returns the final number of examples that will have been
encountered in this run. (The name is slightly non-obvious, but
getTotalCount was already taken.) This allows formatters to report
total run progress as a ratio or a percentage, or display a filling
progress bar, etc.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -201,6 +201,7 @@ runSpec_ config spec = do
   let formatter = fromMaybe specdoc (configFormatter config)
       seed = (fromJust . configQuickCheckSeed) config
       qcArgs = configQuickCheckArgs config
+      numberOfItems = countSpecItems filteredSpec
 
   concurrentJobs <- case configConcurrentJobs config of
     Nothing -> getDefaultConcurrentJobs
@@ -218,6 +219,7 @@ runSpec_ config spec = do
       , formatConfigHtmlOutput = configHtmlOutput config
       , formatConfigPrintCpuTime = configPrintCpuTime config
       , formatConfigUsedSeed =  seed
+      , formatConfigFinalCount = numberOfItems
       }
       evalConfig = EvalConfig {
         evalConfigFormat = formatterToFormat formatter formatConfig
@@ -309,3 +311,6 @@ randomizeForest :: Integer -> [Tree c a] -> [Tree c a]
 randomizeForest seed t = runST $ do
   ref <- newSTRef (mkStdGen $ fromIntegral seed)
   shuffleForest ref t
+
+countSpecItems :: [EvalTree] -> Int
+countSpecItems = getSum . foldMap (foldMap . const $ Sum 1)
--- a/hspec-core/test/Test/Hspec/Core/FormattersSpec.hs
+++ b/hspec-core/test/Test/Hspec/Core/FormattersSpec.hs
@@ -65,6 +65,7 @@ environment = Environment {
   environmentGetSuccessCount = return 0
 , environmentGetPendingCount = return 0
 , environmentGetFailMessages = return []
+, environmentGetFinalCount = return 0
 , environmentUsedSeed = return 0
 , environmentGetCPUTime = return Nothing
 , environmentGetRealTime = return 0
@@ -279,6 +280,14 @@ spec = do
 
     context "same as failed_examples" $ do
       failed_examplesSpec formatter
+
+  describe "additional formatter features" $ do
+    describe "getFinalCount" $ do
+      let formatter = H.silent {H.footerFormatter = fmap show H.getFinalCount >>= H.writeLine}
+          runSpec = captureLines . H.hspecWithResult H.defaultConfig {H.configFormatter = Just formatter}
+      it "counts examples" $ do
+        result:_ <- runSpec testSpec
+        result `shouldBe` "6"
 
 failed_examplesSpec :: H.Formatter -> Spec
 failed_examplesSpec formatter = do
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -3,6 +3,7 @@
   - Add location information for failing afterAll-hooks
   - Remove `--verbose` option (this has been a noop since at least 2013)
   - Remove `--out` option (use shell output redirection instead)
+  - Add `getFinalCount` to `Formatter` API
 
 ## Changes in 2.7.10
   - Add a new formatter (can be used with `--format checks`)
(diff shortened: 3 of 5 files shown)


hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -309,3 +310,6 @@ randomizeForest :: Integer -> [Tree c a] -> [Tree c a]
 randomizeForest seed t = runST $ do
   ref <- newSTRef (mkStdGen $ fromIntegral seed)
   shuffleForest ref t
+
+countExamplesInSpec :: [EvalTree] -> Int
+countExamplesInSpec = getSum . foldMap (foldMap . const $ Sum 1)
```
sol, line 315: We use "example" in many places mostly for historic reasons.  For new code I prefer "spec item" instead.  So please rename to e.g. `countSpecItems` and `numberOfItems`.

hspec-core/test/Test/Hspec/Core/FormattersSpec.hs
```diff
@@ -280,6 +281,14 @@ spec = do
     context "same as failed_examples" $ do
       failed_examplesSpec formatter
 
+  describe "additional formatter features" $ do
+    describe "getFinalCount" $ do
+      let formatter = H.silent {footerFormatter = fmap show getFinalCount >>= writeLine}
+          runSpec = captureLines . H.hspecWithResult H.defaultConfig {H.configFormatter = Just formatter}
+      it "counts examples" $ do
+        result:_ <- runSpec testSpec
+        result `shouldBe` "6"
```
sol, line 290: Actually, I think we are indirectly testing `countExamplesInSpec` here.  Remove this and add a test for `countExamplesInSpec` instead?

sol (changes requested):
I think this is a reasonable addition to the API.  I certainly like it.

I agree that `getFinalCount` is slightly non-obvious, but for a lack of better ideas I guess I'm fine with it (still, suggestions welcome, anybody?).

I left a couple of comments that I want addressed before merging.  Basically `s/example/spec item/` and it should be good to go.

hspec-core/src/Test/Hspec/Core/Runner.hs
rhendric: What do you prefer for doc comments: be consistent with nearby doc comments that use the term ‘examples’, or use the new ‘spec items’?

sol: "spec items".

We want to adjust existing comments that use "examples" for consistency (PRs welcome).

hspec-core/test/Test/Hspec/Core/FormattersSpec.hs
rhendric: You're correct, and I can certainly add a unit test for `countSpecItems` if you like, but:
1. I think having a simple end-to-end test ensuring the API is wired up correctly does add value, and
2. If this test exists, given how simple `countSpecItems` is, I don't know if a unit test adds much additional value. If anything, I would spend my next testing budget dollar on adding more end-to-end tests with more complex specs (with hooks, with more/fewer `describe` sections, etc.).

What do you think?

rhendric:
Do you have a preference for whether this level of code review feedback work comes in as additional commits or as amending the original commit?

sol:
> additional commits or as amending

amending please

hspec-core/test/Test/Hspec/Core/FormattersSpec.hs
sol: > I think having a simple end-to-end test ensuring the API is wired up correctly does add value

I'm ok with that.  So lets keep it as is for now.

sol:
Regarding the test failure, I kind of assumed that this should happen (or rather, I was surprised that we didn't see this error before).

At the point we iterate over the whole spec we force the spec tree and have to keep the whole thing in memory.  So we are gone loose constant space execution.

Not sure yet what to do about it.  I still think this is a reasonable addition to the API.

@nh2 I think you originally cared about running Hspec with constant space.  Any thoughts?

rhendric:
How's this? By making the field on `FormatConfig` strict, I think I've traded space for a little extra time (to generate the filtered tree twice, instead of generating it once and sharing it).

sol:
Ok, looks good now :)

sol:
> I think I've traded space for a little extra time (to generate the filtered tree twice, instead of generating it once and sharing it).

I'm not convinced that this is the case.  Specifically, I think the filtered tree is only generated once.

First let's look at the spec that resulted in the memory blowup (< 1.5m vs ~300m):

```haskell
spec :: Spec
spec = do
  describe "hspec" $ do
    it "does not hold a reference to example action" $ do
      let n = (maxBound `div` 1024) :: Word32
      forM_ [1 .. n] $ \x -> do
        when (x == 0) $ expectationFailure "failed"
```

So here is what I assume happens (until proven wrong that is):

1. At the point we determine `numberOfItems` the tree is partially evaluated to the point needed (specifically, the IO action is not run yet and that huge list inside the IO action is not generated).
2. Later, when we run the spec, we fully evaluate the tree, including that huge list inside the IO action.

With the strictness annotation GHC knows that the IO action is only run once and and garbage collects the lists as it goes over it.

Without the strictness annotation, GHC doesn't know if the IO action will be run again and holds a reference to the fully evaluated list for sharing.

It still bugs me that GHC tries to share that list at all, it doesn't feel intuitive to me, but I guess this is something I just have to accept.  I can't pin it down with some specific set of rules, but intuitively I would assume that the list should not be shared. If somebody can shed some light on this, I'm all ears.

sol:
> it doesn't feel intuitive to me

It turn out that this is indeed unintuitive, but rational.  So why is that?  Let's consider this two claims separately:

# 1. Why is it rational?

Let's express the IO action from above as a one-liner first:
```
action :: IO ()
action = forM_ [1 :: Int .. 5000000] $ (`when` return ()) . (== 0)
```
If we now look at it without caring about the specific types and operators involved we have something of the form:

```haskell
action = f [1 :: Int .. 5000000] `op` x
```
So if we assume that in an expression like `f (23 + 42) + x` the operand to `f` is only evaluated once, then it makes perfect sense that `[1 .. 5000000]` is shared too.

# 2. Why is it still unintuitive then, you ask?

I guess it's just due to the fact that with monadic code we oftentimes don't get sharing.

If we have at least one do binding **before** the *value in question*, say

```haskell
action :: IO ()
action = do
  r <- return ()
  print r
  forM_ [1 :: Int .. 5000000] $ \ x ->
    when (x == 0) $ return ()
```
we loose sharing (for the obvious reason that after desugaring we end up with a lambda before the value in question).

PR #527, opened 2021-05-15
hspec-discover: Add support for custom configs

sol:


hspec-discover/integration-test/with-config/SpecConfig.hs
```diff
@@ -0,0 +1,9 @@
+module SpecConfig where
+
+import Test.Hspec.Runner
+import Test.Hspec.Core.Formatters.V2
+
+config :: Config -> IO Config
+config c = return c {
+  configFormat = Just $ formatterToFormat progress
+}
```
sol, line 9: @soenkehahn can I get your thoughts on this?

This will allow to hard-code custom config options into an `hspec-discover` generated test driver.

`SpecConfig` has to exposes `config :: Config -> IO Config`.  For now the generated test driver will always pass `defaultConfig` as an argument to `config`.

Do you think the type `Config -> IO Config` is over engineered at this point?  Technically `config :: IO Config` (or even `config :: Config`) would work too.

What ever we do here, it will be hard to change in the future (if we ever want to change it, we would probably need to switch to a different module name, say `SpecConfigV2.hs` or something.

soenkehahn: Interesting. I'm reading up on `Config` a bit. If I understand this PR right, it allows `hspec` users to put a `SpecConfig.hs` file into the test suite and put configuration options in there.

[readConfig](https://www.stackage.org/haddock/lts-17.11/hspec-core-2.7.8/Test-Hspec-Core-Runner.html#v:readConfig) has this precedence ordering for the different sources of configuration options. At first glance I would assume that the configuration options from a `SpecConfig.hs` file would be part of that precedence list. I think I would find this ordering the most natural:

1. ~/.hspec (a config file in the user's home directory)
2. .hspec (a config file in the current working directory)
3. `SpecConfig.hs`
4. the environment variable HSPEC_OPTIONS
5. the provided list of command-line options (the second argument to readConfig)

But it seems the current implementation just ignores all other sources of configuration values when a `SpecConfig.hs` is present. Which feels very surprising to me. Is that correct? Is that intended?

(If `SpecConfig.hs` should be in that precedence list, then the type of `SpecConfig.config` would have to be either `Config -> IO Config` or `Config -> Config`.)

sol: The precedence is:

1. `SpecConfig.hs`
1. `~/.hspec` (a config file in the user's home directory)
1. `.hspec` (a config file in the current working directory)
1. the environment variable `HSPEC_OPTIONS`
1. the provided list of command-line options

The rational for this is that `SpecConfig.hs` is per project** configuration, while the others are typically per user.  So to give a user the final say, per user takes precedence over per project.

Does this make sense?

** that is it applies to a group of people working on a project

soenkehahn: Ah, I see. Thanks for explaining. I am a bit surprised by that though. I definitely would have expected a config file in a project to override a config file in `$HOME`. That's also what e.g. `git` does, for what it's worth: https://stackoverflow.com/questions/17756753/where-do-the-settings-in-my-git-configuration-come-from. (I can see why you might want to be able to override project settings with a user config in `~/.hspec`. I just think that precedence would be very surprising.)

Another question: I would've thought that the `.hspec` in the current directory would also be used per project (in the sense you described) most of the time. Since that file would probably be in the project's root directory and it'd be easy and natural to add it to version control. What use-cases are there that can be solved with `SpecConfig.hs`, but not with an `.hspec` config file in the repo?

But to answer your initial question whether I think `Config -> IO Config` is over-engineered: Maybe a little bit, but:

1. I think it's still pretty easy to understand from the type what the user is expected to do. So the cognitive overhead is not that big.
2. Having to introduce a `SpecConfigV2.hs` later would be pretty ugly, I think. Given that there might be more config sources later, I think the `Config` argument is a good idea.
3. Having `IO` might be a good idea, since it actually gives users more power over an `.hspec` file in the project.

I don't feel strongly though. Also, I don't think I've ever used config files for `hspec`, only command line options. So take all of this with a grain of salt. I hope it helps, though!

sol: > That's also what e.g. `git` does

Do you mean `.git/config`?  I'm not sure if this is a fair comparison.  You can't put `.git/config` under revision control, while I assume that `test/SpecConfig.hs` will commonly be put under revision control.

The problem I'm trying to solve is this.  Let's assume I write a custom formatter for Hspec and publish it as a third-party package to Hackage.  Let's call it `my-fancy-format`.  If I want to use `my-fancy-format`, I have to do two things:

1. add `my-fancy-format` to my list of dependencies (`package.yaml` or `*.cabal`).
2. tell Hspec about it

For (2), I can pass the `--format` option to `hspec-discover` (which is deprecated and I'm going to remove it at some point).

The other option I have right now is passing `--module-name` to `hspec-discover` and create a custom test driver like so:

```haskell
-- test/Spec.hs
{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
```
```haskell
-- test/Main.hs
module Main (main) where

import Test.Hspec.Runner
import qualified Spec
import qualified My.Fancy.Format as Fancy

main :: IO ()
main = hspecWith defaultConfig {configFormat = Just Fancy.format} Spec.spec
````

With support for `SpecConfig.hs` I don't have to touch `test/Spec.hs` + I don't have to create a custom test driver, I can create

```haskell
-- test/SpecConfig.hs
module SpecConfig where

import Test.Hspec.Runner
import qualified My.Fancy.Format as Fancy

config :: Config -> IO Config
config c = return c {configFormat = Just Fancy.format}
```
instead.

I think which format to use is a user decision.  Ideally a user could change it in `.hspec`, but there is no easy way to support that (at least I'm not aware of anything, the best I could come up with is this proposal from a couple of years back, which would require support from build tools: https://github.com/sol/haskell-plugins#a-collaborative-approach-to-haskell-development-plugins)

As for `.git/config`, I think the Hspec counterpart is `.hspec` (in the current directory).

sol:
@soenkehahn #555 implements an alternative approach that allows to transform the `Config` within `SpecM`.  With his you can e.g. write something like:

```haskell
spec :: Spec
spec = do
  useFormatter checks
  ...
```

There are two approaches that I'm aware of:
1. Maintain a function `Config -> Config` in `SpecM`
2. Maintain a `Config` value in `SpecM`

(1) is extremely simple to implement.  We just need to change the `WritterT` value from `[SpecTree]` to `(Endo Config, [SpecTree])` and do some plumbing here and there, that's what #555 implements.  (2) would be more powerful, as it allows you to look at the config while constructing the spec tree.

However, both (1) and (2) have one major issue.  Let's look at the definition of `hspec`:

```haskell
hspec :: Spec -> IO ()
hspec spec =
      getArgs
  >>= readConfig defaultConfig
  >>= runSpec spec
  >>= evaluateSummary
```

It's composed of the primitives:
```haskell
readConfig :: Config -> [String] -> IO Config 
runSpec :: Spec -> Config -> IO Summary
evaluateSummary :: Summary -> IO ()
```

If a user needs to customize something it's pretty easy by mixing & matching these primitives.

Now with #555 the `spec` value produces some function `f :: Config -> Config` that we want to apply to the config at some point.  Right now #555 does it in `runSpec`.  This allows us to keep the types of the basic primitives simple and retain the composability of these primitives.  The issue with this is that now config settings from the `spec` precedence over command-line options, which I don't think is acceptable.

If we want to give command-line options precedence then I'm aware of two options:

1. Evaluate the `spec` twice, once in `readConfig` and once in `runSpec`.  However, this is observable, as it will run any IO used withing `spec` twice.
2. Evaluate the `spec` once before `readConfig` at the cost of more complicated types composability.

Not particularly excited about either option.

PR #542, opened 2021-09-10
changed: show -> ushow closed #384

### LINKED ISSUE #384: Human-readable output of unicode characters in expectation results
When there are non-ascii chars in strings, the resulting output shows codes of those chars instead of their human-readable representation, e.g.

```
expected: "\1084\1072\1075\1072\1079\1080\1085"
  but got: "\1052\1072\1075\1072\1079\1080\1085"
```

In GHCi there is a way to specify a *different show function*, e.g. `:set -interactive-print Text.Show.Unicode.uprint`.

Is there a way to similarly alter Hspec's output function to properly handle unicode chars?

ncaq:
relation: [Human-readable output of unicode characters in expectation results · Issue #384 · hspec/hspec](https://github.com/hspec/hspec/issues/384)

Show promotes string escaping, so it is not used in unnecessary parts.

This will result in the following display improvements.

> To rerun use: --match "/FooBar/\30333\12356\29356\12364\36208\12427\12290/"

to

> To rerun use: --match "/FooBar/白い犬が走る。/"

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
+++ b/hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
@@ -71,6 +71,7 @@ import           Test.Hspec.Core.Util
 import           Test.Hspec.Core.Clock
 import           Test.Hspec.Core.Spec (Location(..))
 import           Text.Printf
+import           Text.Show.Unicode (ushow)
 import           Control.Monad.IO.Class
 import           Control.Exception
 
@@ -162,11 +163,11 @@ checks = specdoc {
 
         times
           | dt == 0 = ""
-          | otherwise = " (" ++ show dt ++ "ms)"
+          | otherwise = " (" ++ ushow dt ++ "ms)"
 
     formatProgress (current, total)
-      | total == 0 = show current
-      | otherwise  = show current ++ "/" ++ show total
+      | total == 0 = ushow current
+      | otherwise  = ushow current ++ "/" ++ ushow total
 
 specdoc :: Formatter
 specdoc = silent {
@@ -192,7 +193,7 @@ specdoc = silent {
         writeLine $ indentationFor ("" : nesting) ++ "# PENDING: " ++ fromMaybe "No reason given" reason
       Failure {} -> withFailColor $ do
         n <- getFailCount
-        writeResult nesting (requirement ++ " FAILED [" ++ show n ++ "]") duration info
+        writeResult nesting (requirement ++ " FAILED [" ++ ushow n ++ "]") duration info
 
 , formatterDone = defaultFailedFormatter >> defaultFooter
 } where
@@ -209,11 +210,11 @@ specdoc = silent {
 
         times
           | dt == 0 = ""
-          | otherwise = " (" ++ show dt ++ "ms)"
+          | otherwise = " (" ++ ushow dt ++ "ms)"
 
     formatProgress (current, total)
-      | total == 0 = show current
-      | otherwise  = show current ++ "/" ++ show total
+      | total == 0 = ushow current
+      | otherwise  = ushow current ++ "/" ++ ushow total
 
 progress :: Formatter
 progress = failed_examples {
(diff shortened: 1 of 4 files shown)


hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
```diff
@@ -293,7 +293,7 @@ defaultFailedFormatter = do
         Error _ e -> withFailColor . indent $ (("uncaught exception: " ++) . formatException) e
 
       writeLine ""
-      writeLine ("  To rerun use: --match " ++ show (joinPath path))
+      writeLine ("  To rerun use: --match " ++ ['"'] ++ joinPath path ++ ['"'])
```
sol, line 296: I think this will not behave sane if  `joinPath path` contains `"`-characters.

ncaq: That's right.
I did it this way because I didn't want to add too many dependencies, but it seems that unicode-show is still needed.

ncaq: I added unicode-show.

hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
```diff
@@ -293,7 +294,7 @@ defaultFailedFormatter = do
         Error _ e -> withFailColor . indent $ (("uncaught exception: " ++) . formatException) e
 
       writeLine ""
-      writeLine ("  To rerun use: --match " ++ show (joinPath path))
+      writeLine ("  To rerun use: --match " ++ ushow (joinPath path))
```
sol, line 297: From what I understand, this is the only change that makes an actual difference.  @ncaq is that the case?  Can we revert the other changes?

sol (changes requested):
CI is currently not passing.  I think it's an issue with Cabal version bounds on `unicode-show`.  In addition, from what I understand, `unicode-show` is unmaintained.  To resolve both of these issue, I think we can "vendor" `unicode-show` (read: add the source file + spec to `hspec-core`).  @ncaq can you look into this?

ncaq:
@sol
The author of unicode-show has passed away, but the Japanese community has taken over the maintenance, and the development has continued since then.
I understand that the reason why the last commit is old is simply because there is not much to do, so there is no need to create a new feature.
As for the dependencies, I will try to see if I can fix them here.

hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
ncaq: I decided to use ushow in the formatting module because I thought that escape sequences are nothing to be happy about appearing as a result of formatting.
If you have a strong desire to keep the show, we can put it back.

ncaq:
@sol
The unicode-show depends only on base, and requires 4.7 or higher.
This makes it unusable on older GHCs and cancels all other actions.
I believe that too old GHCs can no longer be removed from the support of the library to be updated. This is because if someone wants to use the old GHC, they can use the old hspec.
However, if there is a strong desire to support older GHCs, I will talk to the maintainer of unicode-show and see if we can lower the required base version of unicode-show.

Translated with www.DeepL.com/Translator (free version)

hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
sol: Yes, please, let's keep `show` except for when `ushow` is needed.

sol:
@ncaq we want to support GHC 7.4.1 and later.

ncaq:
@sol
I fixing unicode-show.

[ncaq/unicode-show: A Haskell Package for unescaping unicode characters in print and show.](https://github.com/ncaq/unicode-show/)

I changed base lower bound of unicode-show to GHC 7.6.

However, it seems to be difficult to make it compatible with 7.4.
I've broken through the problems such as ReadP not being Applicative, but I don't think I can control the problem of Text.Read not providing readMaybe in a safe way.
I suppose I could just copy and paste the readEither implementation, but it's an unsafe function and I don't want to fork it on my own.
Do we really need support for 7.4, or can we please give up on 7.6 somehow?

[fixed: ReadPがApplicativeではなくMonadではある場合に対応する · ncaq/unicode-show@92f8797](https://github.com/ncaq/unicode-show/actions/runs/1443528973)

ncaq:
If I support GHC 7.4, I use CPP, ushow use when GHC 7.6 and leter.
However CPP is ugly, I don't want use.

sol:
@ncaq I don't want to drop support for GHC 7.4 at this time.  I'm closing this in favor of https://github.com/hspec/hspec/pull/572.

PR #546, opened 2021-11-06
Do not depend on clock

sol:
This is trading correctness for fewer dependencies.  One immediate
benefit of this change is that we will get GHC 9.2.1 support right now.

The new code uses getCurrentTime, which is not monotonic.  As a
consequence the new behavior is less accurate when:

- A test run spans a leap second
- The system clock is adjusted during a test run

Both of these events are reasonably rare, so I hope this is the right
thing to do.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Clock.hs
+++ b/hspec-core/src/Test/Hspec/Core/Clock.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE CPP #-}
 module Test.Hspec.Core.Clock (
   Seconds(..)
 , toMilliseconds
@@ -13,9 +14,14 @@ import           Prelude ()
 import           Test.Hspec.Core.Compat
 
 import           Text.Printf
-import           System.Clock
 import           Control.Concurrent
 import qualified System.Timeout as System
+
+#if MIN_VERSION_base(4,11,0)
+import qualified GHC.Clock as GHC
+#else
+import           Data.Time.Clock.POSIX
+#endif
 
 newtype Seconds = Seconds Double
   deriving (Eq, Show, Ord, Num, Fractional, PrintfArg)
@@ -27,9 +33,13 @@ toMicroseconds :: Seconds -> Int
 toMicroseconds (Seconds s) = floor (s * 1000000)
 
 getMonotonicTime :: IO Seconds
+#if MIN_VERSION_base(4,11,0)
+getMonotonicTime = Seconds <$> GHC.getMonotonicTime
+#else
 getMonotonicTime = do
-  t <- getTime Monotonic
-  return $ Seconds ((fromIntegral . toNanoSecs $ t) / 1000000000)
+  t <- getPOSIXTime
+  return $ Seconds (realToFrac t)
+#endif
 
 measure :: IO a -> IO (Seconds, a)
 measure action = do
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -1,3 +1,6 @@
+## Changes in next
+  - Do not depend on `clock`
+
 ## Changes in 2.10.1
   - Support for GHC 9.4.1
   - Generated modules now include an export list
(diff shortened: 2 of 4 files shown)


hspec-core/src/Test/Hspec/Core/Clock.hs
```diff
@@ -28,8 +28,8 @@ toMicroseconds (Seconds s) = floor (s * 1000000)
 
 getMonotonicTime :: IO Seconds
 getMonotonicTime = do
```
soenkehahn, line 30: Maybe this shouldn't be called `getMonotonicTime` anymore, since this is not monotonic anymore. Maybe `getTime`?

soenkehahn:
> Both of these events are reasonably rare, so I hope this is the right thing to do.

Is `getMonotonicTime` only used to measure the speed of tests and display times to the user? If yes, this change doesn't seem bad or dangerous to me.

I'm struggling a bit to understand why this is necessary to support `ghc-9.2` however. Locally I tried to run the `hspec-core` test-suite both from current branch `main` and branch `ghc-9.2.1`, both with `ghc-9.2.1` as the compiler. And both times it builds fine, but the test-suite fails. Here's is the failure message:

```
  test/Test/Hspec/Core/Example/LocationSpec.hs:42:29:
  1) Test.Hspec.Core.Example.Location.extractLocation, with pattern match failure in do expression, in IO, extracts Location
       expected: Just (Location {locationFile = "test/Test/Hspec/Core/Example/LocationSpec.hs", locationLine = 40, locationColumn = 13})
        but got: Nothing
```

Did you run the tests from branch `ghc-9.2.1` with compiler `ghc-9.2.1` successfully? Any idea why it doesn't work on my machine? And why does `clock` need to be replaced?

sol:
This is the error I'm getting with GHC 9.2.1:

```
linking dist/build/System/Clock_hsc_make.o failed (exit code 1)
rsp file was: "dist/build/System/hsc2hscall441822-2.rsp"
command was: /usr/bin/gcc dist/build/System/Clock_hsc_make.o dist/build/System/Clock_hsc_utils.o -o dist/build/System/Clock_hsc_make -fuse-ld=gold -fuse-ld=gold -L/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/base-4.16.0.0 -Wl,-R,/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/base-4.16.0.0 -L/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/ghc-bignum-1.2 -Wl,-R,/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/ghc-bignum-1.2 -lgmp -L/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/ghc-prim-0.8.0 -Wl,-R,/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/ghc-prim-0.8.0 -lc -lm -L/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/rts -Wl,-R,/home/sol/.install/haskell/ghc-9.2.1/lib/ghc-9.2.1/rts -lm -lrt -ldl -lnuma -lpthread
error: /usr/bin/ld.gold: error: cannot find -lnuma
```

I guess this is not `clock` specific, but will probably happen with any package that contains C sources?

Installing `libnuma-dev` resolves the issue.

Not entirely sure if this is a bug? Should e.g. `./configure` fail when installing https://downloads.haskell.org/~ghc/9.2.1/ghc-9.2.1-x86_64-deb10-linux.tar.xz?

Also: https://gitlab.haskell.org/ghc/ghc/-/issues/15444#note_388256

hspec-core/src/Test/Hspec/Core/Clock.hs
sol: done

sol:
> > Both of these events are reasonably rare, so I hope this is the right thing to do.
> 
> Is `getMonotonicTime` only used to measure the speed of tests and display times to the user? If yes, this change doesn't seem bad or dangerous to me.

Yes, exactly.

> Did you run the tests from branch `ghc-9.2.1` with compiler `ghc-9.2.1` successfully? Any idea why it doesn't work on my machine? And why does `clock` need to be replaced?

https://github.com/hspec/hspec/pull/549 addresses the failing test.

sol:
It turns out that we can use `GHC.Clock.getMonotonicTime` with recent versions of GHC.

PR #565, opened 2021-11-10
Hspec discover configs

sol:
Alternative to #555.  Original discussion at https://github.com/hspec/hspec/pull/527.


sol:
Alternative to #555.  Original discussion at https://github.com/hspec/hspec/pull/527.

PR #572, opened 2021-11-15
Preserve unicode in `To rerun use: --match ...` output

sol:


sol:
@ncaq does this work for you?

ncaq:
It work, maybe.
I had a very hard time writing test code that supports GHC 7 series in unicode-show body and got frustrated. (Because there is no readEither, readMaybe).
So, if you choose not to include the test code, I think you can simply run it.
However, it means that it does not contain any test code.

ncaq:
If you don't use the package, you'll have to maintain it on the hspec side, which is more work, but I think it's good that the problem is solved.

PR #573, opened 2021-11-16
Recover unicode from ExpectedButGot (close #384)

### LINKED ISSUE #384: Human-readable output of unicode characters in expectation results

sol:
This turns e.g.

    expected: "\956"
     but got: "\955"

into

    expected: "μ"
     but got: "λ"

### DIFF
--- a/CHANGES.markdown
+++ b/CHANGES.markdown
@@ -1,4 +1,5 @@
 ## Changes in next
+  - Recover unicode from `ExpectedButGot`
   - Preserve unicode in `To rerun use: --match ...` output
 
 ## Changes in 2.9.0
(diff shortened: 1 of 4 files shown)


ncaq:
It great.

PR #575, opened 2021-11-17
Fix glitch?

fisx:
I haven't tested this change, but it makes a lot more sense to me.

### DIFF
--- a/doc/hspec-discover.md
+++ b/doc/hspec-discover.md
@@ -85,7 +85,7 @@ default formatter:
 
 ```haskell
 -- file test/Spec.hs
-{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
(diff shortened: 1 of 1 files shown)


doc/hspec-discover.md
```diff
@@ -85,7 +85,7 @@ default formatter:
 
 ```haskell
 -- file test/Spec.hs
-{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
+{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Main #-}
```
sol, line 88: `Main` is the default.  Reason we want `Spec` here is so that we can import it in `test/Main.hs`: https://github.com/hspec/hspec/pull/575/files#diff-53407afa08299dfbea2617b6878c65a93a7b2d5bb5e197d98ccfd0f152afa695R97

@fisx does this make sense?

fisx:
ah.  right, got it.  sorry for the noise.

PR #588, opened 2021-11-26
Get rid of vendored unicode-show

sol:


hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Unicode.hs
```diff
@@ -0,0 +1,28 @@
...
+, ushows
+) where
+
+import           Prelude ()
+import           Test.Hspec.Core.Compat
+
+import           Data.Char
+
+ushow :: String -> String
+ushow xs = ushows xs ""
```
sol, line 12: @igrep @ncaq if you only care about `String`s, `ushow` can be implemented like this (which is simpler and probably faster).

igrep: Thank you. Fantastic!

Issue #363, opened 2018-07-08
Hspec shows rerun-note for afterAll hooks


sol:
e.g.:

```
Failures:

  1) afterAll-hook
       ...

  To rerun use: --match "/afterAll-hook/"
```

PR #604, opened 2021-12-15
Do not treat afterAll as another spec item (fixes #364, #363)

### LINKED ISSUE #364: Failures in afterAll hooks are included in the sumary

### LINKED ISSUE #414: afterAll seems to behave as another spec item for other hooks
I wanted to have tests that use both `around_` and `afterAll_` hooks. However, the `afterAll_` causes an extra, unexpected run of the `around_` hooks. Here's an example:

```haskell
import Test.Hspec

main :: IO ()
main = hspec $ do
  describe "Prelude.head" $
    around_ (\action -> putStrLn "Around before" >> action >> putStrLn "Around after") $
      afterAll_ (putStrLn "After all") $
        it "returns the first element of a list" $ do
          head [23 ..] `shouldBe` (23 :: Int)
```

The output is

```
Prelude.head
Around before
Around after
  returns the first element of a list
Around before
After all
Around after
```

This is quite confusing because the documentation says that `around_` will "Run a custom action before and/or after every spec item", and there should only be one spec item here, not two.

This seems slightly related to #363 and #364 in the sense that `afterAll` hooks are being treated like spec items.

sol:


hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
```diff
@@ -156,13 +158,59 @@ cancelMany asyncs = do
...
+mapHead :: (a -> a) -> NonEmpty a -> NonEmpty a
+mapHead f xs = case xs of
+  y :| ys -> f y :| ys
+
+addCleanupToItem :: Maybe (String, Location) -> IO () -> RunningItem -> RunningItem
+addCleanupToItem loc cleanup item = item {
+  itemAction = \ path -> do
+    (t1, r1) <- itemAction item path
+    (t2, r2) <- measure $ safeEvaluateResultStatus (cleanup >> return Success)
+    let t = t1 + t2
```
sol, line 196: @parsonsmatt we could refrain from adding the cleanup time here. This would mean that we exclude the times four `afterAll*`. For `aroundAll*` it would skip the *release*, but it will still include the *acquire* portion.

For `before`/`around`/`after` and `befoeAll` it’s not possible to skip the times. Doing so would require conceptional changes.

Issue #346, opened 2018-01-24
Make colourful output per line


ip1981:
hspec should set colour right before printing colourful line and reset it right after.

Currently it seems output multiple lines with colour and this breaks wrappers like timestamps in Jenkins (timestamps become colourful too).

![hspec-color](https://user-images.githubusercontent.com/131844/35326800-76d02b34-0100-11e8-935b-1120da121a18.png)

ip1981:
For some reason github fails to upload a file: https://paste.ec/paste/OYdyDcTE#t9sv5d9MWxUh9WMeMxld+v8cFuUMjBz++3Q47nM2ztP

robx:
Ran into the same issue on buildkite.

<img width="470" alt="Screenshot 2022-01-12 at 18 49 32" src="https://user-images.githubusercontent.com/132113/149221330-9390f39e-a4a0-447f-88d6-3ddd2564c6ff.png">

I made #605 to address this, though I'm not really sure it's entirely comprehensive.

In that screenshot above, not only does the color bleed into the timestamp prefixes. Buildkite also doesn't pick up on the `--- ` line prefix which it usually uses to segment the logs. While #605 ensures that there's no longer a "color reset" after the final newline, we'll still have the output end with `\n\ESC[?25h`, since hspec wraps its colored output in `cursor off`, `cursor on` globally.

Ideally, output would end with a `\n`, but is that really something hspec should need to worry about, or is it rather a limitation in these tools?

sol:
> we'll still have the output end with `\n\ESC[?25h`, since hspec wraps its colored output in `cursor off`, `cursor on` globally.

@robx I addressed this for GitHub actions in 1ee36756 (disabling both progress reporting and cursor hiding).  I'm happy to do something similar for Buildkite.

sol:
I opened https://github.com/hspec/hspec/pull/644.

sol:
This will be available in `hspec-2.9.6` which is about to be released (pending CI, https://github.com/hspec/hspec/pull/651).

sol:
It turns out that in diff output we sometimes need to colorize line breaks: #714, #660

PR #605, opened 2022-01-12
Ensure color codes don't span newlines

robx:
This is meant to address #346. It's not at all polished, but should be technically "done". Do let me know if you agree with this change and the approach in principle -- I'd be more than happy to make any desired changes to make the code nicer.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs
+++ b/hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs
@@ -145,7 +145,8 @@ data FormatterState = FormatterState {
 , stateCpuStartTime    :: Maybe Integer
 , stateStartTime       :: Seconds
 , stateTransientOutput :: String
-, stateConfig :: FormatConfig
+, stateConfig          :: FormatConfig
+, stateColor           :: Maybe SGR
 }
 
 getConfig :: (FormatConfig -> a) -> FormatM a
@@ -167,7 +168,7 @@ runFormatM :: FormatConfig -> FormatM a -> IO a
 runFormatM config (FormatM action) = do
   time <- getMonotonicTime
   cpuTime <- if (formatConfigPrintCpuTime config) then Just <$> CPUTime.getCPUTime else pure Nothing
-  st <- newIORef (FormatterState 0 0 [] cpuTime time "" config)
+  st <- newIORef (FormatterState 0 0 [] cpuTime time "" config Nothing)
   evalStateT action st
 
 -- | Increase the counter for successful examples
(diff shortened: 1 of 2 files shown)


hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs
```diff
@@ -225,9 +226,27 @@ clearTransientOutput = do
 
 -- | Append some output to the report.
 write :: String -> FormatM ()
-write s = do
+write str = do
+  useColor <- getConfig formatConfigUseColor
```
robx, line 230: an alternative would be to check `useColor` in `withColor_`, not sure which is nicer

sol:
@robx this looks great!  I'll rebase + resolve conflicts and merge.

Sorry for the delay.

hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs
sol: I think that is what I would do.  I may refactor it.

hspec-core/src/Test/Hspec/Core/Formatters/Internal.hs
```diff
@@ -259,21 +278,12 @@ htmlSpan :: String -> FormatM a -> FormatM a
...
-      (when useColor $ hSetSGR h [Reset])
-
-      -- run action
-      (runStateT action st)
+withColor_ color action = do
+  oldColor <- gets stateColor
+  modify (\state -> state { stateColor = Just color })
+  x <- action
+  modify (\state -> state { stateColor = oldColor })
+  return x
```
sol, line 286: I really like this.  It's great how we get arbitrary color nesting for free with this approach :clap:

sol:
@robx I squashed your changes into a single commit (to facilitate `git bisect` if needed at some point in the future).  I preserved your authorship, but I changed the commit message to match the title of this PR. I also applied some refactoring on top of it:

1. Move the `useColor` check to `withColor_` https://github.com/hspec/hspec/pull/643/commits/8c4b40796c34b5f94289984ad0db34df958b4cda
2. Only colorize non-whitespace chunks https://github.com/hspec/hspec/pull/643/commits/d76e3efc0874b961cc4b37f1236010ffc7469210
3. With (2) in place, it felt natural to first chunk the output string and the do a `mapM_` over it (instead of `doLines`) https://github.com/hspec/hspec/pull/643/commits/3a4cff2e2c418b394d91267040a0d2a47fde97c7
4. Having unit tests for (3) I thought it's ok to get rid of the test in `RunnerSpec` and instead only test that `write` works for one simple case (we already know that chunking works correctly after all) https://github.com/hspec/hspec/pull/643/commits/9c1b47be3c26ad396c024239407906f3ec8f5cf3

The result is on https://github.com/hspec/hspec/pull/643.

I think this is pretty awesome now.  I really like the approach you took.  Somehow I lacked inspiration on how to tackle this.  Thanks a bunch!  Really much appreciated!

PR #632, opened 2022-03-29
github action formatter

avieth:
Use the option `-f github` to get the typical `V2.checks` formatter but with github actions `::error` commands written out after each failing test. Run this in a github action and it will annotate the action logs as well as inline.

### DIFF
--- a/hspec-core/hspec-core.cabal
+++ b/hspec-core/hspec-core.cabal
@@ -68,6 +68,7 @@ library
       Test.Hspec.Core.Runner
       Test.Hspec.Core.Format
       Test.Hspec.Core.Formatters
+      Test.Hspec.Core.Formatters.GithubAction
       Test.Hspec.Core.Formatters.V1
       Test.Hspec.Core.Formatters.V2
       Test.Hspec.Core.QuickCheck
--- a/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
+++ b/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
@@ -22,6 +22,7 @@ import           Test.Hspec.Core.Compat
 
 import           Test.Hspec.Core.Example (Params(..), defaultParams)
 import           Test.Hspec.Core.Format (Format, FormatConfig)
+import qualified Test.Hspec.Core.Formatters.GithubAction as GA
 import qualified Test.Hspec.Core.Formatters.V1 as V1
 import qualified Test.Hspec.Core.Formatters.V2 as V2
 import           Test.Hspec.Core.Util
@@ -103,6 +104,7 @@ defaultConfig = Config {
   , ("progress", V2.progress)
   , ("failed-examples", V2.failed_examples)
   , ("silent", V2.silent)
+  , ("github", GA.withGithubActionFormatter V2.checks)
   ]
 , configFormat = Nothing
 , configFormatter = Nothing
--- a/hspec-core/src/Test/Hspec/Core/Formatters/GithubAction.hs
+++ b/hspec-core/src/Test/Hspec/Core/Formatters/GithubAction.hs
@@ -1,0 +1,119 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Test.Hspec.Core.Formatters.GithubAction (
+  withGithubActionFormatter
+) where
+
+import Test.Hspec.Core.Format
+import Test.Hspec.Core.Formatters.Internal
+import Test.Hspec.Core.Util (joinPath)
+
+data ErrorCommand = ErrorCommand
+  { title   :: !(Maybe String)
+  , file    :: !(Maybe String)
+  , line    :: !(Maybe Int)
+  , col     :: !(Maybe Int)
+  , message :: !String
+  }
+
+-- | Make a suitable error annotation from an hspec failure.
+--
+-- Not clear what to do with the Maybe Location here: do we use the one from
+-- the Item, or this one? What if both or neither are available?
+-- Also not clear whether to use itemInfo.
+errorCommandFor :: Path -> Item -> Maybe Location -> FailureReason -> ErrorCommand
+errorCommandFor specPath@(_nesting, requirement) item _mLoc reason = ErrorCommand
+    { -- requirement is used because it should always be non-empty and meaningful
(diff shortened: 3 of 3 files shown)


sol (commented):
This looks great. I'm puzzled if we want to enable this by default for GitHub Actions?

parsonsmatt (approved):
Having seen this in action on the test repo, I'm highly in favor! Will make the GitHub Action experience much nicer.

@sol enabling by default on GHA would be a nice UX improvement "for free" for anyone that upgrades. I'd be in favor.

sol:
## Options

I tried a couple of things, going forward I think there are technically 3 options:

1. Enable it by default for the `checks` formatter (and possibly others)
2. Have a separate `github` formatter in `hspec-core` (this PR)
3. Provided it as a separate package

I'm not too eager to do (2) at this point.  Basically because once we add it, we will have to maintain it indefinitely.  This leaves us with (1) or (3).

## Preferred Option: 1+

Initially, my preferred approach is (1) with a minor modification.  Instead of emitting both, the current `Failures:`-list and `::error`-commands, enhance the `Failure:`-list.


Specifically, when `GITHUB_ACTIONS=true`, instead of emitting
```
Data.String.Strip
  strip
    removes leading and trailing whitespace [✘]

Failures:

  test/Data/String/StripSpec.hs:10:7:
  1) Data.String.Strip.strip removes leading and trailing whitespace
       expected: "foo bar baz"
        but got: "foo bar"
```
emit
```
Data.String.Strip
  strip
    removes leading and trailing whitespace [✘]

Failures:

  test/Data/String/StripSpec.hs:10:7:
::error file=test/Data/String/StripSpec.hs,line=10,col=7::/Data.String.Strip/strip/removes leading and trailing whitespace/%0Aexpected: "foo bar baz"%0A but got: "foo bar"%0A
```


Lets call this option (1+).

# Limitations of Option (1+)

Going with (1+) means that we want to colorize things within `::error` (e.g. diffs).  From [what I tried](https://github.com/hspec/hspec-example/runs/5803717073), this kind of works:

screen

However looking at the [summary](https://github.com/hspec/hspec-example/actions/runs/2084188908)

screen

or the [annotated diff](https://github.com/hspec/hspec-example/commit/aae0d8960432c561cae1081cbd83f40668310703#annotation_3127344075)

screen2

we see uninterpreted escape sequences.

As I understand it, we need to compromise somewhere, say choose between:

a) Have redundancy in the build log (having both the current `Failures:`-list with colored diffs + `::error...` without colored diffs)
b) Only go with `::error..` and compromise on color diffs
c) Compromise on the summary / annotated diff, accepting that we have uninterpreted escape sequences

As for (a), I'm not sure.  If a build fails I want to see the error as quickly as possible in a representation that I can parse as quick as possible.  On the one hand, having the errors in the summary/diff view could help with that, on the other hand I might want to see a colorized diff and end up scrolling through the log anyway, except for, now the log is longer, being cluttered with redundant `::error`s.  So it's not a clear win.

Not eager on (b), as I think colorized diffs help a lot with certain types of errors.

If we only care about functionality, then I think (c) is our best bet.  But it looks broken, so :-1:.

# Going forward
I think we should dogfeed option (a) for a while, to see if this is just a novelty, or if it improves productivity in a meaningful way (looking nice alone is not good enough, at least not when it comes at a cost).  I think the best way to do this is, to go with (3) (create a separate package and decide later if we want to fold it into `hspec` proper).

Right now I'm working hard on getting `2.10.0` out.  Once `2.10.0` is released, using third-party functionality with `hspec-discover` will require less boilerplate (https://github.com/hspec/hspec/pull/641).  So I think this will come in handy with dogfeeding.

# Other options
In a perfect world, GitHub Actions would allow us to colorize stuff within `::error:..` in a way that does not screw up the summary / diff annotations.  I think it's worth a try to contact GitHub support about this.  I'm not doing this myself right now, as my priority is on `2.10.0`.  @parsonsmatt @avieth if one of you guys wants to step in, that will be great!

parsonsmatt:
I'd like to get us back upstream. Right now we're stuck on this commit for `hspec-core`, which is blocking other upgrades. If we could get this merged, that'd be awesome. If you don't want the potential maintenance burden, then we can do option 3 with this. Do you have a pointer on making an external package for a formatter like this?

sol:
> Do you have a pointer on making an external package for a formatter like this?

@parsonsmatt you are using `hspec-discover`, right?

Make sure that you read and understand https://hspec.github.io/hspec-discover.html#spec-hooks.

Start with:

```haskell
module SpecHook where

import Test.Hspec
import Test.Hspec.Api.Formatters.V3

withGithubActionFormatter :: Formatter -> Formatter
withGithubActionFormatter formatter = formatter {
  formatterItemDone = \ path item -> do
    formatterItemDone formatter path item
    emitGithubActionAnnotation path item
}

emitGithubActionAnnotation :: Path -> Item -> FormatM ()
emitGithubActionAnnotation path item = write $ show (itemLocation item) -- adjust this to your liking

github :: (String, Formatter)
github = ("github", withGithubActionFormatter checks)

hook :: Spec -> Spec
hook spec = modifyConfig (useFormatter github) >> spec
```

Also read https://hspec.github.io/extending-hspec-formatter.html.  It describes how to implement a formatter from scratch using `Test.Hspec.Api.Format.V1`, so it's not exactly the same thing, but some of it is still transferable.

parsonsmatt:
Great, thank you so much!

We are using `hspec-discover`, but we have our own `TestMain` that does some preprocessing for us.

sol:
> If we could get this merged, that'd be awesome.

I laid out my perspective in detail above. But again, briefly, before upstreaming this, this are the things I want to see:
1. Contact GitHub support and see if they have something regarding ANSI sequences in error annotations on their internal road map, or if this is something they would consider
2. External package so that we can dogfeed it / experiment with it first.

sol:
> we have our own `TestMain` that does some preprocessing for us.

Move it into a `beforeAll_` in `test/SpecHook.hs` if you can.

parsonsmatt:
Ok, cool! I think we are going to go with the external package approach then. Feels like the easiest way to get it integrated. Thanks for the work in making the formatting extensible!!

PR #635, opened 2022-03-31
Assume --color by default when running on GitHub Actions

sol:
Use `HSPEC_COLOR=no` or `--no-color` to opt out of this.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -366,18 +366,24 @@ withHiddenCursor useColor h
 
 colorOutputSupported :: ColorMode -> IO Bool -> IO Bool
 colorOutputSupported mode isTerminalDevice = case mode of
-  ColorAuto  -> (&&) <$> (not <$> noColor) <*> isTerminalDevice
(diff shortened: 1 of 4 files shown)


sol:
@avieth @parsonsmatt On *GitHub* Actions it's not possible to detect whether `stdout` goes to a terminal or is redirected to a file.  For that reason Hspec's `auto` behavior does not work on GitHub and resorts to `--no-color`.  As a consequence most people will not see colors on GitHub (unless they specifically opt in via e.g. `HSPEC_COLOR=yes`).

[sol/run-haskell-tests](https://github.com/sol/run-haskell-tests) takes care of it already (indeed, it's one of the two reasons that motivated me to create this action in the first place, [Cabal's broke `test` behavior](https://github.com/haskell/cabal/issues/5079) being the other).

But as it's already so convoluted to "correctly" run Haskell tests on CI, I guess any simplification is a good thing.

The obvious downside is that if you don't want colors (e.g. when redirecting to a file), you have to say so explicitly (e.g. `--no-color` or `HSPECE_COLOR=no`).

Thoughts?

parsonsmatt:
This seems reasonable to me!

PR #636, opened 2022-03-31
Add a nix shell environment

srid:
This PR doesn't fully nixify the project (I assume no active maintainer uses Nix), but it does add a minimal nix-shell setup for potential contributors, like me, who use Nix on a daily basis and do not have Haskell tools installed globally.

See https://nixos.wiki/wiki/Development_environment_with_nix-shell

The `flake.nix` should require little to no maintenance over time.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1648297722,
+        "narHash": "sha256-W+qlPsiZd8F3XkzXOzAoR+mpFqzm3ekQkJNa+PIh1BQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "0f8662f1319ad6abf89b3380dd2722369fc51ade",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1648097358,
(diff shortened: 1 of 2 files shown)


sol:
Heads up, this is unlikely to land.

@srid is there anything Hspec specific in that setup, or is this boilerplate that you would more or less reproduce verbatim for every Haskell project you work on?

> The `flake.nix` should require little to no maintenance over time.

Does this lock nixpkgs?  You would still need to update that sometimes, no?

srid:
> @srid is there anything Hspec specific in that setup,

No.

> or is this boilerplate that you would more or less reproduce verbatim for every Haskell project you work on?

Technically true, but if a `flake.nix` is not committed to Git, then `nix develop` won't recognize it. I actually did not intend to open this PR - however, being forced to commit `flake.nix` in my branch was what made me do.

> Does this lock nixpkgs? 

Yes.

> You would still need to update that sometimes, no?

At infrequent intervals, yes - but it is straightforward to do as indicated by the comment above `nixpkgs.url`.

sol:
> > @srid is there anything Hspec specific in that setup,
> 
> No.

I'm not using Nix on a regular basis, but if I were I would want to address this (either for my personal setup, or for the nix community as a whole) in a way so that it works for all my Haskell projects without the need of adding repetitive boilerplate to each and every project that I want to work on.

From what I understand, adding these nix expressions to every project is inherently unDRY.  Let's assume some fancy new LSP implementation comes along that I want to use.  With this approach I would have to updated the nix expression in every single project instead of at one authoritative place.

If somebody things my basic assessment is flawed in some way then please try to convince me otherwise.

PR #644, opened 2022-04-01
Disable progress reporting on Buildkite

sol:


sol:
@robx this should disable cursor hiding on Buildkite.  I don't have a setup to test it.  Can you give it a try?

Also, do colors work out of the box on Buildkite or do you explicitly need to enable them via `--color`?

sol:
This will be available in `hspec-2.9.6` which is about to be released (pending CI, https://github.com/hspec/hspec/pull/651).

PR #648, opened 2022-04-04
Add `getSpecDescriptionPath`

### LINKED ISSUE #637: Add a way to get ancestor test group path in SpecM
```haskell
newtype SpecM a r = SpecM (WriterT (Endo Config, [SpecTree a]) (ReaderT Env IO) r)
  deriving (Functor, Applicative, Monad)
```

This adds an `Env` that stores the "current" test group ancestor that can be retrieved in any `describe` block:

```haskell
-- | Environment of a test definition, in the Spec monad
data Env = Env 
  { envAncestorGroups :: [String]  -- Should this be non empty?
  }
  deriving (Eq, Show, Ord)

askAncestors :: SpecM a [String]
askAncestors = do 
  SpecM $ lift $ asks envAncestorGroups
```

I need `askAncestors` in order to determine the paths to golden files. The idea is to determine this path automatically depending on where in the test hierarchy one is.

---

Is this the right approach to solve this? I can open a PR.

srid:
Resolves #637

I recommend squash merging this PR. Otherwise, let me know if I should clean up commit history.


hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -66,7 +72,11 @@ runIO :: IO r -> SpecM a r
 runIO = SpecM . liftIO
 
 mapSpecForest :: ([SpecTree a] -> [SpecTree b]) -> SpecM a r -> SpecM b r
-mapSpecForest f (SpecM specs) = SpecM (mapWriterT (fmap (fmap (second f))) specs)
+mapSpecForest = mapSpecForest_ id
+
+mapSpecForest_ :: (Env -> Env) -> ([SpecTree a] -> [SpecTree b]) -> SpecM a r -> SpecM b r
+mapSpecForest_ ef tf (SpecM specs) = SpecM $ flip mapWriterT specs $ 
+  mapReaderT (fmap (fmap (second tf))) . withReaderT ef
```
srid, line 79: I'm not happy with this naming. Because `mapSpecItem_ = mapSpecItem id` - and the underscore variant here is flipped. I kept it as is for backwards compatibility on `mapSpecForest`, though.

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -10,19 +10,25 @@ module Test.Hspec.Core.Spec.Monad (
...
 
 , mapSpecForest
+, mapSpecForest_
 , mapSpecItem
 , mapSpecItem_
 , modifyParams
 
 , modifyConfig
+
+, Env(..)
```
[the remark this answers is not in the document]
srid: Oh, I could just put it in a separate module I suppose.

hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -62,10 +62,11 @@ import           Test.Hspec.Core.Example
 import           Test.Hspec.Core.Hooks
 import           Test.Hspec.Core.Tree
 import           Test.Hspec.Core.Spec.Monad
+import           Test.Hspec.Core.Spec.Env (pushAncestor)
 
 -- | The @describe@ function combines a list of specs into a larger spec.
 describe :: HasCallStack => String -> SpecWith a -> SpecWith a
-describe label = mapSpecForest (return . specGroup label)
+describe label = mapSpecForest_ (pushAncestor label) (return . specGroup label)
```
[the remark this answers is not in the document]
sol: Away from the computer, but I guess it can’t… we would still

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
srid: This became redundant using the `local` refactor.

srid:
@sol What's a good place to add tests for this? I tried `SpecSpec.hs`, but that appears to be using a different monad type.

image

hspec-core/src/Test/Hspec/Core/Spec.hs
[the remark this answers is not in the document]
srid: I've done this in c1e2269df257414314639b795163cd584297fece

srid:
Just pushed tests. I think this should be ready for another review.

hspec-core/test/Test/Hspec/Core/SpecSpec.hs
```diff
@@ -20,6 +20,30 @@ spec :: Spec
...
+              ancs2 `shouldBe` ["grandchild1", "child1"]
+          H.describe "grandchild2" $ do
+            ancs2 <- H.askAncestors
+            H.it "works with two-level tree" $ do
+              ancs2 `shouldBe` ["grandchild2", "child1"]
+        H.describe "child2" $ do
+          ancs1 <- H.askAncestors
+          H.it "works with one-level tree" $ do
+            ancs1 `shouldBe` ["child2"]
+      last res `shouldBe` "5 examples, 0 failures"
```
sol, line 45: Most of the work is done by `ReaderT` and `local`. I think if we got the basic wiring  right it’s very hard to have a bug here. What I think is, one simple test case will be enough. I think we can also test it directly, without using `runSpec`, something like:

```
void . runSpecM $ do
  describe …
    describe…
      askAncestors >>= runIO . (`shouldBe` […])
```

hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -29,7 +29,18 @@ module Test.Hspec.Core.Spec (
...
+, Test.Hspec.Core.Spec.Monad.SpecM(..)
+, Test.Hspec.Core.Spec.Monad.runSpecM
+, Test.Hspec.Core.Spec.Monad.fromSpecList
+, Test.Hspec.Core.Spec.Monad.runIO
+, Test.Hspec.Core.Spec.Monad.mapSpecForest
+, Test.Hspec.Core.Spec.Monad.mapSpecItem
+, Test.Hspec.Core.Spec.Monad.mapSpecItem_
+, Test.Hspec.Core.Spec.Monad.modifyParams
+, Test.Hspec.Core.Spec.Monad.modifyConfig
+, Test.Hspec.Core.Spec.Monad.askAncestors
```
sol, line 43: 👍

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -15,14 +16,22 @@ module Test.Hspec.Core.Spec.Monad (
 , modifyParams
 
 , modifyConfig
+
+, askAncestors
+-- END RE-EXPORTED from Test.Hspec.Core.Spec
+
+, mapEnv
+, Env(..)
```
sol, line 24: 👍

hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -65,7 +76,9 @@ import           Test.Hspec.Core.Spec.Monad
 
 -- | The @describe@ function combines a list of specs into a larger spec.
 describe :: HasCallStack => String -> SpecWith a -> SpecWith a
-describe label = mapSpecForest (return . specGroup label)
+describe label = mapSpecForest (return . specGroup label) . mapEnv pushAncestor
+  where 
+    pushAncestor (Env ancs) = Env $ label : ancs
```
sol, line 81: 👍

sol (commented):
Code wise this looks great. So time for bike shedding. `sydtest`'s naming (`getTestDescriptionPath`) while lengthy, is more obvious to me. How about `getSpecPath`, or alternatively `getSpecDescriptionPath`?

hspec-core/test/Test/Hspec/Core/SpecSpec.hs
srid: I didn't know about `runIO`; looks like it is something to use when evaluating a test without `it`/`describe`? I've reduced the tests.

srid:
I've no opinion on the naming, but `getSpecDescriptionPath` -- being explicit -- does sound better. "description path" returns the path trail left by `describe` groups. So it makes sense.

hspec-core/test/Test/Hspec/Core/SpecSpec.hs
sol: Not exactly. `runIO` runs some IO during spec construction time. This is why we can use it to verify that `getSpecDescriptionPath` returns the right thing during spec construction.

sol (changes requested):
I think from a users perspective this looks good now :+1: 

- [ ] For consistency, rename `envAncestorGroups` to `envSpecDescriptionPath`
- [ ] Update `CHANGES.markdown`
- [ ] Squash the commits into a single commit and make sure the commit message is descriptive

@srid if you are around and wanna do it then please go ahead.  Otherwise I might go ahead with it.

PR #655, opened 2022-04-07
Reverse the list of labels returned by `getSpecDescriptionPath`

sol:


hspec-core/src/Test/Hspec/Core/Spec.hs
```diff
@@ -185,8 +185,18 @@ pending_ = (E.throwIO (Pending Nothing Nothing))
...
+--   describe "foo" $ do
+--     describe "bar" $ do
+--       getSpecDescriptionPath >>= runIO . print
+-- :}
+-- ["foo","bar"]
 --
 -- @since 2.10.0
 getSpecDescriptionPath :: SpecM a [String]
-getSpecDescriptionPath = SpecM $ lift $ asks envSpecDescriptionPath
+getSpecDescriptionPath = SpecM $ lift $ reverse <$> asks envSpecDescriptionPath
```
sol, line 202: @srid I'm inclined to return the list in reversed order (for consistency with the `Path` type, used e.g. by the formatters).  Does this work for you?

srid: 👍🏾

PR #667, opened 2022-04-18
Allow to transform the execution environment of a spec item

sol:


srid:
@sol I've wanted this ability as well, so 👍🏿 - I'm curious what prompted it in your case (see the linked PRs for why I think we need this).

hspec-core/example/CustomMoandStackSpec.hs
```diff
@@ -0,0 +1,54 @@
+{-# LANGUAGE DerivingStrategies #-}
```
L-as, line 1: FWIW there's a typo in the file name.

sol:
> I've wanted this ability as well, so 👍🏿
@srid it will take a while until I'll be able to get back to this, but quickly, what do you need:

1. Access additional configuration in `SpecM`?
2. Access additional configuration from within hooks / an `it` / a custom `Example` instance?

srid:
@sol More like,

1. *Add* project-specific custom test configuration type, and
1. *Modify* that configuration for a certain sub-tree (`describe`) of spec forest

I thought the `StateT Int` in your example demonstrated pretty much this (assuming `Int` is that configuration). In our case, this configuration would be a configuration type specifying where to generate golden tests and which goldens, in particular, to create (of which certain tests may override).

sol:
@srid that sounds like (1), right?  This PR is (2).  However, I think (1) will be easier.  There are some complications with this PR that I'll try to note down when I have time.  But in short, the approach in this PR does not play nice with how we are doing QC properties.  So it's not clear if/when this will land (possibly we will have it as an alternate approach for experimentation).

@srid if you want your use case covered, can you please open a separate issue and give a minimal use case (actually code with imports that type checks, fill in the things that are not possible with `hspec` right now with `undefined`).

PR #674, opened 2022-04-26
Make exception handling of `aroundAll` more robust

sol:


parsonsmatt (commented):
Do you mind doing a self-review and calling out what you think are important features of this PR? Stuff like, "What's the problem behavior we're trying to resolve?"

sol (commented):
@parsonsmatt Hey, I was unavailable for the last 2 weeks and tried to get this finished before that but didn't manage to provide any context for the change.

I added a couple of comments that hopefully fix that.

For the new code I think it makes sense to look at the whole function (say "View file") instead of mentally piecing it together from the diff).

I think the tests provide reasonable confidence that this works on a basic level.  I'm not aware of anything, but just to be sure, can you double check that everything is sane in regards to async exceptions / masking?

Regarding (1), this doesn't matter with how we use the code right now (that's the reason why it didn't surface), but it will make a difference with #673.  So technically this is a preparatory commit for #673.  But let's ignore #673 for now, as I think this PR is an improvement even without #673.

PR #680, opened 2022-06-17
[hspec-discover] Add --parallel option

blackheaven:


sol:
I think this is already possible:

```haskell
-- test/SpecHook.hs
module SpecHook where

import Test.Hspec

hook :: Spec -> Spec
hook = parallel
```

@blackheaven does this work for you?

We probably want to add this to the documentation.

blackheaven:
Looks great, yeah, documenting it would be great

PR #682, opened 2022-06-24
Call release only when acquired in hooks

### LINKED ISSUE #679: Tests blocked indefinitely with nested failed `aroundAll`
There seems to be an edge case for hspec 2.10.0 in which a failing `aroundAll` call can result in a frozen test execution. It only happens with hspec 2.10.0. Version 2.9.7 doesn't seem affected by this.

### Steps to reproduce

Run the given code with hspec 2.10.0:
```haskell
import Control.Exception
import Test.Hspec

data TestException = SomeError
    deriving (Show)

instance Exception TestException

main :: IO ()
main = hspec $
    aroundAll actionF $ do
        describe "Some test group inside a failing around" $ do
            describe "Some valid tests" $ do
                it "succeeds on its own but fails because of the failing around" True
            aroundAll actionS $ do
                describe "Some test inside a successing around but it doesn't matter" $ do
                    it "succeeds on its own but fails because of the failing root around" True
  where
    actionS e = e ()
    actionF e = throwIO SomeError
```

We get the following output:
```text
Some test group inside a failing around
  Some valid tests
    succeeds on its own but fails because of the failing around [✘]
  Some test inside a successing around but it doesn't matter
hspec-issue-test: thread blocked indefinitely in an MVar operation      

hspec-issue> Test suite hspec-issue-test failed
Completed 2 action(s).
Test suite failure for package hspec-issue-0.1.0.0
    hspec-issue-test:  exited with: ExitFailure 1
Logs printed to console
```

Note that the `hspec-issue-test: thread blocked indefinitely in an MVar operation` appears randomly. Though this basic example does finish after some time, we originally encountered the issue in a much larger project. Its test run actually always freezes and we have to Ctrl-C out of it. I could not reproduce this freezing part, but the `thread blocked indefinitely in an MVar operation` message was pretty close so hopefully the root cause is the same.

I would have expected no blocking and some output close to what hspec 2.9.7 is producing:
```text
Some test group inside a failing around
  Some valid tests
    succeeds on its own but fails because of the failing around [✘]
  Some test inside a successing around but it doesn't matter
    succeeds on its own but fails because of the failing root around [‐]
      # PENDING: exception in aroundAll-hook (see previous failure)
  afterAll-hook [‐]
    # PENDING: exception in aroundAll-hook (see previous failure)
afterAll-hook [‐]
  # PENDING: exception in aroundAll-hook (see previous failure)

Failures:

  test/Spec.hs:20:17: 
  1) Some test group inside a failing around, Some valid tests, succeeds on its own but fails because of the failing around
       uncaught exception: TestException
       SomeError

  To rerun use: --match "/Some test group inside a failing around/Some valid tests/succeeds on its own but fails because of the failing around/"

Randomized with seed 241965352

Finished in 0.0016 seconds
4 examples, 1 failure, 3 pending
```

Have I missed something? Thanks in advance for your help

michivi:
This PR fixes an issue when the `release` action in `aroundAll_` and `aroundAllWith` is called when the `acquire` action was not, causing a deadlock.

This situation can happen with multiple `aroundAll_` / `aroundAllWith` when one of the earlier `acquire` action has failed but the `release` actions of uncalled `acquire` still get executed.

Fixes #679.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Hooks.hs
+++ b/hspec-core/src/Test/Hspec/Core/Hooks.hs
@@ -2,6 +2,9 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ConstraintKinds #-}
+#if __GLASGOW_HASKELL__ < 710
+{-# LANGUAGE DeriveDataTypeable #-}
+#endif
 -- | Stability: provisional
 module Test.Hspec.Core.Hooks (
   before
--- a/hspec-core/src/Test/Hspec/Core/Hooks.hs
+++ b/hspec-core/src/Test/Hspec/Core/Hooks.hs
@@ -33,6 +33,7 @@ module Test.Hspec.Core.Hooks (
 import           Prelude ()
 import           Test.Hspec.Core.Compat
 import           Data.CallStack (HasCallStack)
+import           Data.Typeable (Typeable)
 
 import           Control.Exception (Exception, SomeException, finally, throwIO, try)
 import           Control.Concurrent
@@ -142,7 +143,7 @@ data OrderedAcquireReleaseState
 data OrderedAcquireReleaseException
   = HasAlreadyBeenAcquired
   | HasAlreadyBeenReleased
-  deriving Show
+  deriving (Show, Typeable)
 
 instance Exception OrderedAcquireReleaseException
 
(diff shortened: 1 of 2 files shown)


michivi:
Oh, sorry. I tested this with GHC 8.10.7 only. The `Typeable` instance has been added in https://github.com/hspec/hspec/pull/682/commits/3ce69212c19899655c7e4e0889f103766cabbf97.

sol:
@michivi Hey! Thanks a lot for tracking this down.

I'm inclined to fix this in `decompose` and simplify it to #684.

PR #684, opened 2022-06-29
Simplify "Call release only when acquired in hooks"

sol:
(ass introduced in #682)


sol:
@michivi can you give this a quick review?

michivi:
Looks good to me! And much simpler! Thanks! 😀

PR #687, opened 2022-08-07
Compute diffs (actual, expected) concurrently

blackheaven:
In our code base, during a heavy refactoring, we had the issue having a lot (~250) of test failing quickly (mostly HTTP endpoints) hence the total execution was taking a lot of time (50s to run the tests, in parallel, 10 minutes to have all the results, one by one).

After a bit of code review, it appears that, while tests can be run in parallel, results are computed (especially the diff algorithm) sequentially.

To emphasis my issue I came up with a simple case:

```haskell
module Main where

import Control.Monad
import Data.List
import Test.Hspec

main :: IO ()
main = do
  hspec $
    parallel $ do
      forM_ ([0 .. 10] :: [Int]) $ \_ ->
        it "Quick test" $ do
          shouldBe (show left) (show right)

data S = S
  { t :: String,
    f0 :: Maybe S,
    f1 :: Maybe S,
    f2 :: Maybe S,
    f3 :: Maybe S
  }
  deriving (Eq, Show)

data T = T
  { u :: String,
    g5 :: Maybe T,
    g6 :: Maybe T,
    g7 :: Maybe T,
    h8 :: Maybe T
  }
  deriving (Eq, Show)

left :: Maybe S
left = mkS "azrtqyui"

right :: Maybe T
right = mkT "azmXauryt"

mkS :: String -> Maybe S
mkS x =
  case permutations x of
    (t' : f0' : f1' : f2' : f3' : _) -> Just $ S t' (mkS (tail f0')) (mkS (tail f1')) (mkS (tail f2')) (mkS (tail f3'))
    _ -> Nothing

mkT :: String -> Maybe T
mkT x =
  case permutations x of
    (t' : f0' : f1' : f2' : f3' : _) -> Just $ T t' (mkT (tail f3')) (mkT (tail f2')) (mkT (tail f0')) (mkT (tail f1'))
    _ -> Nothing
```

which, once execute, gives:

```
	Sun Aug  7 10:57 2022 Time and Allocation Profiling Report  (Final)

	   slow +RTS -N20 -p -RTS --times

	total time  =        4.94 secs   (13936 ticks @ 1000 us, 20 processors)
	total alloc = 46,118,315,688 bytes  (excludes profiling overheads)
```

and

```
cabal exec slow -- +RTS -N20 -p -RTS --times  29.8s user 1.94s system 139% cpu 26.3 total
```

My PR, which is still a draft:

```
	Sun Aug  7 15:44 2022 Time and Allocation Profiling Report  (Final)

	   slow +RTS -N20 -p -RTS --times

	total time  =        7.65 secs   (21577 ticks @ 1000 us, 20 processors)
	total alloc = 12,257,225,264 bytes  (excludes profiling overheads)
```

and

```
cabal exec slow -- +RTS -N20 -p -RTS --times  27.78s user 1.94s system 377% cpu 7.873 total
```

I'd like to gather feedback regarding:
* Is it relevant to be integrated into hspec?
* what would be a better design? (for the moment it's a proof of concept, from the top of my head, I can add a configuration for the parallelism, or compute the diff at the end of each test)
* Anything relative to performance (I barely get a x4 on a 20 threads CPU)

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
+++ b/hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
@@ -82,10 +82,12 @@ import           Data.Maybe
 import           Test.Hspec.Core.Util
 import           Test.Hspec.Core.Clock
 import           Test.Hspec.Core.Example (Location(..))
+import           Test.Hspec.Core.Format(PrettyPrintFunction)
 import           Text.Printf
 import           Test.Hspec.Core.Formatters.Pretty.Unicode (ushow)
 import           Control.Monad.IO.Class
 import           Control.Exception
+import           Control.Concurrent.Async
 
 -- We use an explicit import list for "Test.Hspec.Formatters.Monad", to make
 -- sure, that we only use the public API to implement formatters.
@@ -258,15 +260,34 @@ defaultFailedFormatter = do
     writeLine "Failures:"
     writeLine ""
 
-    forM_ (zip [1..] failures) $ \x -> do
+    computeDiffs <- useDiff
+    precomputedDiffs <-
+      if computeDiffs
+        then do
+          pretty <- prettyPrintFunction
+          liftIO $ mapConcurrently (precomputeDiff pretty) failures
+        else return $ repeat Nothing
+
+    forM_ (zip3 [1..] failures precomputedDiffs) $ \x -> do
       formatFailure x
       writeLine ""
 
     write "Randomized with seed " >> usedSeed >>= writeLine . show
     writeLine ""
   where
-    formatFailure :: (Int, FailureRecord) -> FormatM ()
-    formatFailure (n, FailureRecord mLoc path reason) = do
+    precomputeDiff :: Maybe PrettyPrintFunction -> FailureRecord -> IO (Maybe [Diff])
+    precomputeDiff pretty (FailureRecord _ _ reason) =
+      case reason of
+        ExpectedButGot _ expected_ actual_ -> do
+          let
+            threshold = 2 :: Seconds
+            (expected, actual) = case pretty of
+              Just f -> f expected_ actual_
+              Nothing -> (expected_, actual_)
+          timeout threshold (evaluate $ diff expected actual)
+        _ -> return Nothing
+    formatFailure :: (Int, FailureRecord, Maybe [Diff]) -> FormatM ()
+    formatFailure (n, FailureRecord mLoc path reason, precomputedDiff) = do
       unicode <- outputUnicode
       forM_ mLoc $ \loc -> do
         withInfoColor $ writeLine ("  " ++ formatLocation loc)
@@ -284,15 +305,7 @@ defaultFailedFormatter = do
 
           mapM_ indent preface
 
-          b <- useDiff
-
-          let threshold = 2 :: Seconds
-
-          mchunks <- liftIO $ if b
-            then timeout threshold (evaluate $ diff expected actual)
-            else return Nothing
-
-          case mchunks of
+          case precomputedDiff of
             Just chunks -> do
               writeDiff chunks extraChunk missingChunk
             Nothing -> do
--- a/hspec-core/src/Test/Hspec/Core/Format.hs
+++ b/hspec-core/src/Test/Hspec/Core/Format.hs
@@ -5,6 +5,7 @@
 module Test.Hspec.Core.Format (
   Format
 , FormatConfig(..)
+, PrettyPrintFunction
 , Event(..)
 , Progress
 , Path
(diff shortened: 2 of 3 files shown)


sol:
@blackheaven  sorry for the late reply.

Some thoughts (in no particular order):

- The underlying problem is that the diff implementation is slow, much slower than GNU diff.  Doing things in parallel may alleviate this somewhat, but it does not address the underlying problem.
- https://github.com/hspec/hspec/issues/514 would address the underlying problem, at the cost of less control and likely less useful diff output.

On performance:

- As I understand it, the performance of GHC’s garbage collector may degrade if you give it too many cores (this was at least historically an issue on 32 core Travis CI machines, where test runs would effectively grind to a halt).
- With hyper-threading (educated guess, you are running this on a 10900K) you get a 30% performance increase at best (vs just using the 10 physical cores).
- This is technically a parallel (as opposed to concurrent problem). I’m puzzled if using GHC’s support for parallel computing would give us better results.

Things I’d  like to see:
1. A parallel implementation + performance numbers.
2. We currently have a 2s timeout on the diff calculation. I guess you want to remove that when doing a parallel implementation. If the parallel implementation looks good then we may later apply the 2s timeout to the calculation of the whole list. This way we will stale at most 2s in total (vs the current 2s per failing test) on trying to calculate diffs.
3. A summry / interpretation of any numbers and an easy way to reproduce them (e.g. add your test test setup as a separate commit; we can then drop that commit before merging).

Out of curiosity, I’d be interested in what numbers you get if you decrease the number of cores to 5 or 10. But it has little practical significance. Whatever we do here has to yield sane results by default. We cannot expect an end-user to do any kind of performance tuning before running their test suite.

blackheaven:
> Some thoughts (in no particular order):
> 
>     * The underlying problem is that the diff implementation is slow, much slower than GNU diff.  Doing things in parallel may alleviate this somewhat, but it does not address the underlying problem.
>     * https://github.com/hspec/hspec/issues/514 would address the underlying problem, at the cost of less control and likely less useful diff output.

Actually, I looked at [tree-diff](https://hackage.haskell.org/package/tree-diff) but since it altered the output, I moved forward.


> *    As I understand it, the performance of GHC’s garbage collector may degrade if you give it too many cores (this was at least historically an issue on 32 core Travis CI machines, where test runs would effectively grind to a halt).
> *    With hyper-threading (educated guess, you are running this on a 10900K) you get a 30% performance increase at best (vs just using the 10 physical cores).
> *    This is technically a parallel (as opposed to concurrent problem). I’m puzzled if using GHC’s support for parallel computing would give us better results.

I have an i9-10850K :P 

TBH, it's the first time I had such poor parallelization gain with GHC, my guess is: either I used Async/parellism wrong, or I didn't use it at the right place.

> 2. We currently have a 2s timeout on the diff calculation. I guess you want to remove that when doing a parallel implementation. If the parallel implementation looks good then we may later apply the 2s timeout to the calculation of the whole list. This way we will stale at most 2s in total (vs the current 2s per failing test) on trying to calculate diffs.

Not really, I mean, it seems to still be a valid choice to have per test rendering timeout, even if I can understand how useful it would be to have a global timeout (but which would be more complex).

> Out of curiosity, I’d be interested in what numbers you get if you decrease the number of cores to 5 or 10.

From memory it was not significant (maybe 5-10%).

> We cannot expect an end-user to do any kind of performance tuning before running their test suite.

I agree.

sol:
> Not really, I mean, it seems to still be a valid choice to have per test rendering timeout, even if I can understand how useful it would be to have a global timeout (but which would be more complex).

To clarify, I think GHC’s support for parallel computing only works on pure expressions. So if you want to give that a try then you can’t do the timeout on a per test basis.

blackheaven:
> To clarify, I think GHC’s support for parallel computing only works on pure expressions. So if you want to give that a try then you can’t do the timeout on a per test basis.

That may explain the performance issue: in my PR, IIRC, I use map concurrently (from `Async`), I'll have a look at it.

hspec-core/src/Test/Hspec/Core/Formatters/V2.hs
```diff
@@ -258,15 +260,34 @@ defaultFailedFormatter = do
...
+    precomputeDiff :: Maybe PrettyPrintFunction -> FailureRecord -> IO (Maybe [Diff])
+    precomputeDiff pretty (FailureRecord _ _ reason) =
+      case reason of
+        ExpectedButGot _ expected_ actual_ -> do
+          let
+            threshold = 2 :: Seconds
+            (expected, actual) = case pretty of
+              Just f -> f expected_ actual_
+              Nothing -> (expected_, actual_)
+          timeout threshold (evaluate $ diff expected actual)
```
sol, line 287: I actually don’t see where we would sequence the list that is returned by `diff expected actual`. @blackheaven can you try to sequence the list and possibly also make the `Diff` data type strict? Does this give you better numbers?

Also, when you first experimented with this. did you get full utilization of your 20 cores?

sol: > Also, when you first experimented with this. did you get full utilization of your 20 cores?

Looking at your numbers, I think you didn’t. I think we are not sequencing. Right now I’m puzzled how the `timeout` could possibly work at all.

PR #689, opened 2022-08-11
Avoid possible GHC warning about no export lists

mpilgrem:
`Test.Hspec.Discover.Run.mkSpecModule` makes a module without an export list. Since GHC 8.4.1, GHC ~~has warned~~ can warn about missing export lists ~~(as part of -Wall)~~. This pull request aims to ~~silence~~ avoid that possible warning.


sol:
@mpilgrem  thanks for looking into this. Two thoughts:

1. Do you think we should generate an export list instead?
2. If not, could we disable the warning unconditionally (read: is that flag supported by all versions of GHC that we test on CI?).

sol:
There are also some tests that need to be fixed + we want to add 
  - 9.2.3
  - 9.2.4
  - 9.4.1
  
to CI.

\+ We will need a version bump (use `util/bump-version`).

mpilgrem:
@sol, on your questions:

1. I think the compiler warnings are there to help people coding and reading code, not machines generating code that will not be read by people, so I do not think it is _necessary_ to generate an export list.
2. I understand from the GitHub workflow that you test GHC 7.4.1 onwards. The flag was introduced by GHC with GHC 8.4.1 (see https://downloads.haskell.org/~ghc/8.4.1/docs/html/users_guide/8.4.1-notes.html#compiler) - hence the CPP.

mpilgrem:
On the version bump, I tried (with `hpack version 0.35.0` on the PATH) but the script fails with:
~~~
generated hspec.cabal
hspec-core/package.yaml: UnknownAlias {_anchorName = "version"}
~~~

The problem seems to be that `_: !include version.yaml` in the relevant `package.yaml` files is not including `&version 2.10.1`. The `package.yaml` seems to be looking at the local `version.yaml` (which has `../version.yaml`) and that does not 'redirect' to the `version.yaml` in the project root.

As a workaround, to create updated Cabal files, I forced the content of the local `version.yaml` to be `&version 2.10.1` rather than `../version.yaml`.

mpilgrem:
I think I have identified why I am experiencing the problem. On Linux, files can be shortcuts to other files ('symlinks'). On Windows, Git converts those files to text files containing the destination - see https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows/59761201#59761201.

sol:
Yes, that wouldn't work on Windows I guess. Maybe you would need Windows Subsystem for Linux or something.

hspec-discover/src/Test/Hspec/Discover/Run.hs
```diff
@@ -62,6 +62,11 @@ mkSpecModule src conf nodes =
   ( "{-# LINE 1 " . shows src . " #-}\n"
   . showString "{-# LANGUAGE NoImplicitPrelude #-}\n"
   . showString "{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}\n"
+#ifdef MIN_VERSION_GLASGOW_HASKELL
+#if MIN_VERSION_GLASGOW_HASKELL(8,4,0,0)
+  . showString "{-# OPTIONS_GHC -fno-warn-missing-export-lists #-}\n"
+#endif
+#endif
```
sol, line 69: There are corresponding tests for this code in `RunSpec.hs` that need to be updated.

sol (changes requested):
There are a couple of failing tests that need attention.

Regarding CPP vs export list: What you say about warnings and generated code makes sense.  If we only need CPP at that single place in `Run.hs`, then I’m good. However, if this approach results in CPP being splattered around `RunSpec.hs` too, then I say go with an explicit export list instead.

If you end up with the “explicit export list” route, then make sure to take the `configNoMain` into account.

sol:
@mpilgrem thanks for working on this, btw. This is critical to fix. Really appreciate it and sorry for the hiccups and delays.

Two things:

1. Right now I need to manually approve CI runs for your PRs. I can add you to the organization. That way CI will run right away and you will get quicker feedback.
2. As you are on Windows, any feedback, PRs, etc that improve Hspec in that department are very much welcome.

mpilgrem:
I have removed GHC 9.4.1 from the CI, because Hspec's tests do not pass on GHC 9.4.1 for reasons unrelated to this pull request. Moving Hspec to GHC 9.4.1 will need to be a separate endeavour.

hspec-discover/src/Test/Hspec/Discover/Run.hs
mpilgrem: To avoid CPP, I have gone down the export list route.

mpilgrem:
@sol, I think I have responded to your review and it passes the CI. As noted above, Hspec's tests can't pass on GHC 9.4.1, so I have removed that from the CI and raised a separate issue (#690).

sol:
> `Test.Hspec.Discover.Run.mkSpecModule` makes a module without an export list. Since GHC 8.4.1, GHC has warned about missing export lists (as part of -Wall). This pull request aims to silence that warning.

@mpilgrem Just to clarify, did you mean to write `-Weverything` instead of `-Wall`?

mpilgrem:
@sol, you are right, `-Wall` is a mistake. I misread the GHC User Guide. I'll correct the commit message.

hspec-discover/src/Test/Hspec/Discover/Run.hs
```diff
@@ -62,24 +62,20 @@ mkSpecModule src conf nodes =
...
-              showString "main :: IO ()\n"
-            . showString "main = hspec spec\n"
-          True -> ""
-
-moduleName :: FilePath -> Config -> String
-moduleName src conf = fromMaybe (if configNoMain conf then pathToModule src else "Main") (configModuleName conf)
+ where
+  ifNoMain x y = if configNoMain conf then x else y
+  moduleName = fromMaybe (ifNoMain (pathToModule src) "Main")
+                         (configModuleName conf)
```
sol, line 76: Please don’t align code. Why you ask? If somebody later renames `moduleName` (or adds a parameter) then the alignment will break. This is bad for two reasons:

- It discourages refactoring by making it more expensive
- It produces a larger diff, making rebases harder / conflicts more likely

It’s not a big issue here, but it’s best to not get into the habit of aligning code in the first place. At the point you align record fields, etc, it’s more of an issue.

The short story is, aligned code looks nice when you write it, but if you “design your code for change”, it’s not an effective approach.

hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
```diff
@@ -68,7 +68,7 @@ spec = do
           "{-# LINE 1 \"test/Spec.hs\" #-}"
         , "{-# LANGUAGE NoImplicitPrelude #-}"
         , "{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}"
-        , "module Main where"
+        , "module Main (main, spec) where"
```
sol, line 71: We don’t have test coverage for the configNoMain=True case. Can you add a test case for `--no-main`?

sol (changes requested):
This looks mostly good now. I left an other round of review comments.

We also want an acceptance test in `hspec-discover/integration-test/` that enables the warning so that we don’t accidentally break this feature it in the future.

version.yaml
```diff
@@ -1 +1 @@
-&version 2.10.0.1
+&version 2.10.1
```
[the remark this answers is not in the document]
mpilgrem: I've undone the version bump.

hspec-discover/src/Test/Hspec/Discover/Run.hs
```diff
@@ -62,24 +62,20 @@ mkSpecModule src conf nodes =
...
+  exports = " (" ++ (ifNoMain "" "main, ") ++ "spec)"
+  driver = showString $ ifNoMain "" "main :: IO ()\nmain = hspec spec\n"
```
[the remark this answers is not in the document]
mpilgrem: Now done.

hspec-discover/src/Test/Hspec/Discover/Run.hs
mpilgrem: The second line and indentation/alignment arose because I was trying to avoid long code lines. I've found a way to acheive both objectives.

hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
mpilgrem: I've added a test.

mpilgrem:
On code quality more generally, `hlint hspec-discover\src` makes 8 hints unrelated to the this pull request. Would you be interested in a pull request that puts them through?

sol (approved):
We still want an acceptance test. I’ll take care of that.

sol:
> On code quality more generally, `hlint hspec-discover\src` makes 8 hints unrelated to the this pull request. Would you be interested in a pull request that puts them through?

We would need to look at them individually.  I do not always agree with what `hlint` suggests (with default settings at least).  But it’s likely that there are valid suggestions that I’m happy to integrate.

sol:
@mpilgrem so yes, please open a PR and I’ll take a quick look.

PR #695, opened 2022-08-22
Use ubuntu-20.04 on CI

sol:
This will make it difficult to support GHC `7.6.*` and `7.8.*`.  We may eventually decide to drop support for `GHC < 8`.

https://github.com/actions/runner-images/issues/6002


mpilgrem:
What is HSpec's GHC support philosophy? Perhaps it can be framed as: 'as long as possible but at least X years from the date of release, subject to being no longer than Y years from the date of release'? If Y was 8, you would have an objective criteria for dropping GHC 7.8.4 (released 23 Dec 2014).

sol:
I think for all practical means and purposes it should be fine. The only thing is that, as a testing framework, Hspec also sets the stage for what other packages can support. For that reason I don't needlessly break support.

Issue #389, opened 2018-11-14
Add a flag to fail on pending, focused, etc. tests


sol:
For CI use I want a flag that fails if there are any spec items that

- are pending
- are focused
- have an empty description

Maybe `--strict` or something.

sol:
Only adding `--fail-on-focused` for now.

sol:
We now have `—strict` and `—fail-on=focused,pending,empty` + we have a PR for `--fail-on=empty-description` (#700 ).

Closing.

PR #700, opened 2022-08-24
Add --fail-on=empty-description

sol:


sol:
@parsonsmatt  can I get your input on this?

This is part of #389. We already have:

- `—fail-on-focused`
- `—fail-on-pending`
- `—strict`,  which enables all of them

I have two PRs that add to that list:

- #697
- #700 (this PR)

All of these flags are mostly intended for use on CI. You literally never want to commit focused items. Committing pending items is sometimes tempting, but generally bad practice.

Regarding `—fail-on-empty`, personally I don’t have a huge need for it. But I’m not eager to special case the semantics of `—skip=“”` + having all tests ignored on CI doesn’t sound like something you want in a regular basis. So I guess it doesn’t hurt?

`--fail-on-empty-description` (this PR) is something I want. I don’t think there’s a use case for committing empty descriptions. My concern here is the longish flag name and how it affects the help output. I would go with `—fail-on-empty` instead, but that’s already taken by #697 (assuming we use forward with that).

Not exactly sure what to do here + eager to hear your opinion.

hspec-core/test/Test/Hspec/Core/Config/OptionsSpec.hs
```diff
@@ -115,10 +115,11 @@ spec = do
...
           unlines [
             "my-spec: option `--fail' is ambiguous; could be one of:"
-          , "    --fail-on-focused      fail on focused spec items"
-          , "    --fail-on-pending      fail on pending spec items"
-          , "    --fail-fast            abort on first failure"
-          , "    --failure-report=FILE  read/write a failure report for use with --rerun"
+          , "    --fail-on-focused            fail on focused spec items"
+          , "    --fail-on-pending            fail on pending spec items"
+          , "    --fail-on-empty-description  fail on spec items with empty descriptions"
+          , "    --fail-fast                  abort on first failure"
```
parsonsmatt, line 121: This seems pretty reasonable!

I think my main thought here is that - we've got a bunch of `--fail-on-*` options. Possibly it's time to write `--fail-on` as a flag that can be repeated multiple times with an enum argument, like,

```
my-tests --fail-on first --fail-on focused --fail-on pending --fail-on empty-description
```

This reduces the surface area of new CLI arguments and can make it easier to document the behavior - instead of N CLI flags, you have 1 CLI flag with an enum of N constructors.

parsonsmatt (approved):
One comment on overall CLI API design, but that can be punted to a future issue easily

Looks great and seems very reasonable :+1:

hspec-core/test/Test/Hspec/Core/Config/OptionsSpec.hs
sol: I like that idea. Maybe even allow comma separated lists, e.g.
```
spec --fail-on=pending,focused,empty-description
```
We should give that a try eventually.

PR #716, opened 2022-09-06
Suppress excessive diff output

### LINKED ISSUE #449: Suppress excessive diff output
https://github.com/hspec/hspec-expectations/issues/30

sol:
- [x] Add a flag to disable (`--diff-context=full`)
- [x] Suppress identical lines within the text, not only at the start/end
- [x] Never omit less than 3 lines

(close #449)

#### before:
before
#### after (with #715):
after
#### after (without #715):
after

### DIFF
--- a/changes/716.md
+++ b/changes/716.md
@@ -1,0 +1,1 @@
+  - Suppress excessive diff output (#449)
--- a/hspec-core/help.txt
+++ b/hspec-core/help.txt
@@ -35,6 +35,8 @@ FORMATTER OPTIONS
            --[no-]color            colorize the output
            --[no-]unicode          output unicode
            --[no-]diff             show colorized diffs
+           --diff-context=N        output N lines of diff context (default: 3)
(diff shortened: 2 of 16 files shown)


sol:
@chshersh you originally requested this in https://github.com/hspec/hspec-expectations/issues/30.  If you have any input regarding the output formatting then please let me know.  I'm not sure yet if #715 is essential, or if we can do without it.

chshersh:
Hi @sol, this looks fantastic 👏🏻 I don't have any specific input on the output formatting.

I believe we can do without #715. But to my taste it looks better with less indentation because lines can become quite long and it looks nicer when a single line from the origin doesn't span multiple lines in the test output (assuming having the same width for my source and terminal).

sol:
This is the ready to go. I’m going to deal with #715 as a separate story.

parsonsmatt:
Oh that's really nice 👍🏻

PR #718, opened 2022-09-07
Add --fail-on

sol:


hspec-core/src/Test/Hspec/Core/Config/Definition.hs
```diff
@@ -229,14 +229,41 @@ setMaxShrinks n c = c {configQuickCheckMaxShrinks = Just n}
...
 runnerOptions :: [Option Config]
 runnerOptions = [
     mkFlag "dry-run" setDryRun "pretend that everything passed; don't verify anything"
   , mkFlag "focused-only" setFocusedOnly "do not run anything, unless there are focused spec items"
 
-  , mkFlag "fail-on-focused" setFailOnFocused "fail on focused spec items"
-  , mkFlag "fail-on-pending" setFailOnPending "fail on pending spec items"
-  , mkFlag "strict" setStrict "enable --fail-on-focused and --fail-on-pending"
+  , undocumented $ mkFlag "fail-on-focused" setFailOnFocused "fail on focused spec items"
+  , undocumented $ mkFlag "fail-on-pending" setFailOnPending "fail on pending spec items"
```
parsonsmatt, line 262: Can we put warnings/deprecations on these options? If the intent is migrating to a new style, it'd be good to have guidance and a timeline on that

parsonsmatt (approved):

hspec-core/src/Test/Hspec/Core/Config/Definition.hs
sol: I was just going to soft-deprecate them. I didn’t see any cost in keeping them, except for that they clutter the documentation.

1. We currently don’t have any infrastructure in place to warn on “deprecated” flags.
1. If we had the infrastructure in place, then I would probably be ok with issuing a warning.
   - I’m not going to add that infrastructure myself. Getting to **zero open issues** that’s what’s on my agenda right now. If you care about it enough then please go ahead 😉
   - In any case, we would wait at least one major release cycle before issuing warnings; to make transition smoother when switching between projects that are on different versions of `hspec` (think shared `.hspec` files / environment variables / GitHub actions). 
1. We are not going to remove `—fail-on-focused` any time soon. I have used this all over the place for CI jobs, for clients  + I’m not eager to break other people’s builds.

PR #739, opened 2022-09-12
Allow the use of an external diff program, add `--diff-command`

### LINKED ISSUE #514: Allow the use of an external diff program

sol:
(close #514)

#### Hspec's default diff with `:main`
hspec-diff
#### GNU diff with `:main --diff-command "diff -u --color -L expected -L actual"`
gnu-diff

#### `:main --diff-command "git diff"` with `diff-so-fancy`
git-diff

### DIFF
--- a/changes/514.md
+++ b/changes/514.md
@@ -1,0 +1,2 @@
+  - Allow the use of an external diff program.  This can e.g. be used by
+    setting: `HSPEC_DIFF_COMMAND="diff -u --color -L expected -L actual"`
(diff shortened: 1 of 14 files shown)


parsonsmatt (approved):
looks great!

sol:
@parsonsmatt  thanks for taking a look. I had to jump through some hoops to make CI pass on Windows as [`process` relies on some env vars to find `cmd.exe`](https://github.com/haskell/process/blob/master/System/Process/Windows.hsc#L370).

@snoyberg FYI, because you are active on `process`, this adds a dependency on `process`. If at some point in the future somebody wants to use Hspec for the test suite of process then I'm more than happy to make the dependency optional through a flag or maybe even have a foreign import for the standard C version of `system()` in `hspec-core`.

snoyberg:
Thanks for the heads-up @sol, I avoided using hspec in the process test suite specifically for this reason :)

Issue #599, opened 2021-12-08
hspec-core-2.9.3 fails to build for ghcjs


harendra-kumar:
[This](https://app.circleci.com/pipelines/github/composewell/streamly/3417/workflows/08859c5f-d2a8-419d-a095-2c1c2cd17711/jobs/15549) is the failure build.

```
[11 of 40] Compiling Test.Hspec.Core.Formatters.Pretty.Parser ( src/Test/Hspec/Core/Formatters/Pretty/Parser.hs, dist/build/Test/Hspec/Core/Formatters/Pretty/Parser.js_o )

src/Test/Hspec/Core/Formatters/Pretty/Parser.hs:54:1: error:
    Could not find module ‘Lexer’
    Use -v to see a list of the files searched for.
   |
54 | import           Lexer
   | ^^^^^^^^^^^^^^^^^^^^^^

src/Test/Hspec/Core/Formatters/Pretty/Parser.hs:55:1: error:
    Could not find module ‘Parser’
    Use -v to see a list of the files searched for.
   |
55 | import qualified Parser as GHC
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/Test/Hspec/Core/Formatters/Pretty/Parser.hs:56:1: error:
    Could not find module ‘StringBuffer’
    Use -v to see a list of the files searched for.
   |
56 | import           StringBuffer
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

...

cabal: Failed to build hspec-core-2.9.3 (which is required by hspec-2.9.3)
```

Looks like this is due to the ghc dependency introduced by https://github.com/hspec/hspec/commit/0e9ad052be14c639e7981958524b952906b3ae8a . We can possibly avoid pretty printing on ghcjs if it cannot be made to work.

sol:
@harendra-kumar thanks for reporting this.  #600 is a tentative solution.

One thing that puzzles me is that apparently ghcjs provides the `ghc` package, but with a different API?  If `ghc` would not be available in the first place we could use an "automatic" cabal flag to detect this, which might be more robust.

In addition, will this be an issue with cross-compilation in general?  Say, is `ghc` available when cross-compiling (@angerman, any thoughts)?

sol:
I'm going to release #599 as `2.9.4`.  I'm still not very satisfy for three reasons:

- I don't understand what exactly the `ghc` package provides for GHCJS
- I don't have a test setup, ideally we would have CI for GHCJS
[...]

PR #753, opened 2022-11-25
Make the dependency on the ghc package optional

### LINKED ISSUE #752: hspec-core unconditionally depends on the ghc package which pins down versions of ghc boot libraries
Can you please gate the vendored libraries behind a flag so they don't have to be forced?

arybczak:
Fixes #752.

### DIFF
--- a/hspec-core/hspec-core.cabal
+++ b/hspec-core/hspec-core.cabal
@@ -28,6 +28,11 @@ source-repository head
   location: https://github.com/hspec/hspec
   subdir: hspec-core
 
+flag pretty-printing
+  default: True
+  description: Support for pretty printing of data structures via GHC.
+  manual: False
+
 library
   hs-source-dirs:
       src
@@ -96,7 +101,8 @@ library
       Data.Algorithm.Diff
       Paths_hspec_core
   default-language: Haskell2010
-  if impl(ghc >= 8.2.1)
+  if flag(pretty-printing) && impl(ghc >= 8.2.1) && impl(ghc < 9.5)
+    ghc-options: -DPRETTY_PRINTING_SUPPORTED
     build-depends:
         ghc
       , ghc-boot-th
@@ -217,7 +223,8 @@ test-suite spec
       Test.Hspec.Core.UtilSpec
       Paths_hspec_core
   default-language: Haskell2010
-  if impl(ghc >= 8.2.1)
+  if flag(pretty-printing) && impl(ghc >= 8.2.1) && impl(ghc < 9.5)
+    ghc-options: -DPRETTY_PRINTING_SUPPORTED
     build-depends:
         ghc
       , ghc-boot-th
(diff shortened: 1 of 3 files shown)


hspec-core/hspec-core.cabal
```diff
@@ -28,6 +28,11 @@ source-repository head
   location: https://github.com/hspec/hspec
   subdir: hspec-core
 
+flag pretty-printing
```
sol, line 31: This file is generated. Any changes should go into `package.yaml` instead.

hspec-core/hspec-core.cabal
```diff
@@ -28,6 +28,11 @@ source-repository head
...
+  default: True
+  manual: True
```
sol, line 33: Did you try if an "automatic" flag works instead?

hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Parser.hs
```diff
@@ -16,12 +16,6 @@ import           Prelude ()
 import           Test.Hspec.Core.Compat hiding (fail)
 import           Test.Hspec.Core.Formatters.Pretty.Parser.Types
 
-#ifndef __GHCJS__
```
sol, line 19: I think we want to keep this check (see #599).

hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Parser.hs
```diff
@@ -16,12 +16,6 @@ import           Prelude ()
...
-#if __GLASGOW_HASKELL__ >= 802 && __GLASGOW_HASKELL__ <= 904
```
sol, line 20: I think we want to keep the check for `<= 904` (either here or in `package.yaml`), so that Hspec keeps building  with  upcoming versions of GHC (changes to the GHC API are the norm, not the exception).

sol (changes requested):

hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Parser.hs
arybczak: This can also be expressed in the .cabal file with upper bound in impl(ghc), assuming that hpack makes it expressible.

hspec-core/hspec-core.cabal
arybczak: Does hpack support `impl` conditionals? I can't see it in its readme.

hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Parser.hs
arybczak: Actually I'm not sure this is the case (and CI doesn't seem to include GHCJS), so I'll include this condition just to be safe.

hspec-core/hspec-core.cabal
arybczak: Uh, I was looking at a different `package.yaml`, the one for `hspec-core` already had these conditionals. Fixed.

hspec-core/hspec-core.cabal
arybczak: Indeed now I tried it and it works better, i.e. if the situation from #752 arises, Cabal disables the flag automatically.

hspec-core/src/Test/Hspec/Core/Formatters/Pretty/Parser.hs
arybczak: Yeah sorry, I was so distracted I didn't even read your comment fully. I added appropriate upper bound to `package.yaml`.

PR #764, opened 2022-12-18
Add `xprop` and `fprop`, which use `xit` and `fit`

lehins:
I felt like these are missing. 

Feel free to close this PR if you don't think this helpers are needed.


sol (approved):
This looks great :+1:

PR #800, opened 2023-02-15
RFC draft PR of user defined Slow spec item filter

codygman:
One of the most useful things about the slow item filter for my team was seeing when you personally add a slow test and being able to fix it right then.

The problem is we had over IIRC 300-500 items in our slow log filter. Ideally we'd just fix them all but...

In large projects it can sometimes be difficult or impossible to get rid of all slow tests. 

For instance one of our hspec tests uses [validateEveryToJSON](https://hackage.haskell.org/package/servant-swagger-1.1.11/docs/Servant-Swagger.html#v:validateEveryToJSON) which parses all of our Servant routes and ensures every handler has _compatible_ ToJSON and ToSchema instances.

Without a user defined filter, these auto-generated tests would be 100 entries in the slow item log. With the PR's implemenation we fixed that with a user defined filter like:

```haskell
ignoredSlowSpecItems :: [String]
ignoredSlowSpecItems =
  [ "Servant/Swagger/Internal" ]

runHspecWithSlowQueryLogFilter :: Hspec.Spec -> IO ()
runHspecWithSlowQueryLogFilter = runHspecWithConfig Runner.defaultConfig {Runner.configUserSlowSpecFilter = slowQueryLogFilter}

slowQueryLogFilter :: Runner.SlowItem -> Bool
slowQueryLogFilter slowItem =
  let slowItemPath = Hspec.joinPath (Runner.path slowItem)
      slowItemMinimumDuration :: Int = 300
   in not (any (`List.isInfixOf` slowItemPath) ignoredSlowSpecItems)
        && Runner.duration slowItem > slowItemMinimumDuration
```

This was the simplest change I could think of that didn't seem horrible to address this use case, so very open to feedback and suggestions on implementing it differently.


sol:
I don't have much free cycles at the moment. So very quickly:
- Moving forward, I want to encourage putting more functionality into third-party packages so that experimentation/progress is not blocked on upstream. That's something we could try here.
- From what I understand, ideally, you would want to only have times for tests that you added/changed. Not sure how that could work, but I think e.g. `jest` looks at uncommitted changes to figure out what tests to run. Doing something similar for times could be neat, but possibly much harder/impossible to do reliably.

codygman:
Assuming you mean https://jestjs.io. That idea sounds interesting, I'll have to think more about how it would work though.

sol:
@codygman just to make sure, you are not blocked by this, right?

For the time being you could create your own version of `Test.Hspec.Core.Runner.PrintSlowSpecItems` that does exactly what you want and use that instead of `--print-slow`. Does this work for you?

codygman:
> @codygman just to make sure, you are not blocked by this, right?

Not blocked, just trying to figure out how to depend on less custom code and upstream anything generally useful.

> For the time being you could create your own version of Test.Hspec.Core.Runner.PrintSlowSpecItems that does exactly what you want and use that instead of --print-slow. Does this work for you?

Do you mean in my own codebase or that I should submit a PR adding that module?

PR #803, opened 2023-02-26
Add `paramsUseColor` to `Params` record

sol:


sol:
@parsonsmatt this feels a little bit hackish but not sure what better options there are.  Any thoughts?

hspec-core/src/Test/Hspec/Core/Example.hs
```diff
@@ -53,12 +53,14 @@ class Example e where
 data Params = Params {
   paramsQuickCheckArgs  :: QC.Args
 , paramsSmallCheckDepth :: Maybe Int
+, paramsUseColor :: Bool
```
parsonsmatt, line 56: I *generally* prefer to see more specific types - `Bool`ean blindness can be an issue when reading code. `data UseColor = YesUseColor | Don'tUseColor` or something analogous.

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -368,8 +369,8 @@ runSpecForest_ config spec = do
 
   return results
 
-specToEvalForest :: Config -> [SpecTree ()] -> [EvalTree]
-specToEvalForest config =
+specToEvalForest :: Bool -> Config -> [SpecTree ()] -> [EvalTree]
+specToEvalForest useColor config =
```
parsonsmatt, line 373: And, ya, that's where I get Weird feelings about this - a `Bool` parameter is almost always difficult to reason about unless you remember the specific implementation here.

parsonsmatt (commented):
Looks good to me - is this just to support having configured color output in the hspec-hedgehog?

hspec-core/src/Test/Hspec/Core/Example.hs
sol: > Looks good to me - is this just to support having configured color output in the hspec-hedgehog?

Yes, exactly. As of now I can't get colors from `hspec-hedgehog` when redirecting `stdout`.

I'm mentally toying around with a different approach though. I think technically it's the responsibility of the formatter to decide on when to use color output. I'll ping you when I get around to sketch something up.

sol:
> Looks good to me - is this just to support having configured color output in the hspec-hedgehog?

Yes, exactly. As of now I can't get colors from `hspec-hedgehog` when redirecting `stdout`.

I'm mentally toying around with a different approach though. I think technically it's the responsibility of the formatter to decide on when to use color output. I'll ping you when I get around to sketch something up.

sol:
Closing in favor of https://github.com/hspec/hspec/pull/821.

PR #804, opened 2023-03-03
refactoring: Internally use types that more closely represent the domain

sol:
(in `Test.Hspec.Core.Runner`)


parsonsmatt (approved):
awesome~

PR #805, opened 2023-03-05
Strip ANSI escape sequences on `--no-color` from example output

sol:


sol:
Closing in favor of https://github.com/hspec/hspec/pull/821.

PR #821, opened 2023-04-08
Add `ColorizedReason` constructor to `FailureReason`

sol:


sol:
@parsonsmatt the basic idea here is that `hspec-hedgehog` indiscriminately produces color output. It's then the responsibility of the runner to strip ANSI sequences as needed.

The backward compat story for this gets somewhat messy. If you want to give this a review, then it might make sense to ignore any changes to `hspec-api`, at least initially.

PR #834, opened 2023-04-27
Don't crash when home directories are unsupported

amesgen:
From the [`getHomeDirectory` Haddocks](https://hackage.haskell.org/package/directory-1.3.8.1/docs/System-Directory.html#v:getHomeDirectory):

> The operation may fail with:
>
> - `UnsupportedOperation` The operating system has no notion of home directory.
> - [`isDoesNotExistError`](https://hackage.haskell.org/package/base-4.16.3.0/docs/System-IO-Error.html#v:isDoesNotExistError) The home directory for the current user does not exist, or cannot be found.

The motivating example for this PR is the WASM backend in GHC 9.6, as [WASI](https://wasi.dev/) does not support home directories.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Config.hs
+++ b/hspec-core/src/Test/Hspec/Core/Config.hs
@@ -18,6 +18,7 @@ module Test.Hspec.Core.Config (
 import           Prelude ()
 import           Test.Hspec.Core.Compat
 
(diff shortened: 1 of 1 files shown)


sol (approved):
Thanks @amesgen 🎉

We really want to add WASM to CI.

PR #863, opened 2023-10-02
Run cleanup actions on `--fail-fast` (close #860)

### LINKED ISSUE #860: with fail-fast the cleanup action of aroundAllWith is not executed
I'm running my test suite with `--fail-fast`. When there are multiple spec items in an `aroundAllWith` block, and the first one fails, the cleanup action of the hook is not executed.

- it works with `aroundWith`
- it works when the failing test is the last (or the only) in the block
- it works when running without `--fail-fast`

The following is a minimal example that produces the issue.

* Expected outcome: both "initialize" and "teardown" are printed.
* Actual outcome: only "initialize" is printed.

```
module Main (main) where

import Test.Hspec
import Control.Exception

spec :: Spec
spec =
  aroundAllWith myHook $ do
    it "foos" $
      True `shouldBe` False
    it "bars" $
      True `shouldBe` True

main :: IO ()
main = hspec spec

myHook :: (ActionWith () -> ActionWith ())
myHook unitTests state =
  bracket_
    (putStrLn "initialize")
    (putStrLn "teardown")
    (unitTests state)

```

sol:


hspec-core/src/Test/Hspec/Core/Runner/Eval.hs
```diff
@@ -90,14 +91,15 @@ reportItemStarted = formatEvent . Format.ItemStarted
 reportItemDone :: Path -> Format.Item -> EvalM ()
 reportItemDone path item = do
   let
```
mklinik, line 93: this let is empty

mklinik:
I don't claim to understand how the fix works, but I tried it with my example from the issue and it works.

Issue #338, opened 2017-12-22
Add a way to mark slow tests and excluded them by default


sol:
e.g. by tagging tests or something

dnadales:
Adding my vote, as this will be really helpful for us at [`TorXakis`](https://github.com/TorXakis/TorXakis). CC: @pjljvandelaar.

sol:
I think the following could work as an API for this feature...

```haskell
tag :: String -> SpecWith a -> SpecWith a
tag = undefined
```

...which then can be used like so...

```haskell
it "behaves according to the spec" >>> tag "slow" >>> tag "db" $ do
  ...
```

PR #868, opened 2023-11-22
Add extensions API

sol:
This PR adds an extension API that can be used to extend Hspec with additional functionality without the need to modify `hspec-core`.

An example extension that addresses https://github.com/hspec/hspec/issues/338:

```haskell
module Test.Hspec.CI (use, only) where

import System.Environment

import Test.Hspec.Core.Extension
import Test.Hspec.Core.Extension.Item qualified as Item
import Test.Hspec.Core.Extension.Spec qualified as Spec
import Test.Hspec.Core.Extension.Tree qualified as Tree
import Test.Hspec.Core.Extension.Config qualified as Config

-- |
-- Do not run the spec items of the given subtree by default; execute them only
-- when the environment variable @CI@ is set.
--
-- It is idiomatic to use this in combination with `>>>`:
--
-- @
-- import Control.Arrow ((>>>))
-- import Test.Hspec
-- import qualified Test.Hspec.CI as CI
--
-- spec :: Spec
-- spec = do
--   describe ".." $ do
--     it ".." >>> CI.only $ do
--       ..
-- @
only :: SpecWith a -> SpecWith a
only = Spec.mapItems setCiOnly

data CiOnly = CiOnly

setCiOnly :: Item a -> Item a
setCiOnly = Item.setAnnotation CiOnly

getCiOnly :: Item a -> Maybe CiOnly
getCiOnly = Item.getAnnotation

newtype CiFlag = CI Bool

setCiFlag :: Bool -> Config -> Config
setCiFlag = Config.setAnnotation . CI

getCiFlag :: Config -> CiFlag
getCiFlag = fromMaybe (CI False) . Config.getAnnotation

use :: SpecWith a
use = do
  runIO (lookupEnv "CI") >>= \ case
    Nothing -> pass
    Just _ -> modifyConfig (setCiFlag True)
  registerOption $ flag "ci" setCiFlag "include itmes that are marked as \"CI only\""
  registerTransformation applyCiOnly

applyCiOnly :: Config -> [SpecTree] -> [SpecTree]
applyCiOnly config = case getCiFlag config of
  CI True -> id
  CI False -> Tree.mapItems $ \ item -> case getCiOnly item of
    Nothing -> item
    Just CiOnly -> Item.pendingWith message item
  where
    message = unlines [
        "This item is marked as \"CI only\" and excluded by default."
      , "Use `--ci' to include this item."
      ]
```

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Extension.hs
+++ b/hspec-core/src/Test/Hspec/Core/Extension.hs
@@ -79,7 +79,7 @@ import           Test.Hspec.Core.Compat
 import qualified Data.CallStack as CallStack
 
 import qualified GetOpt.Declarative as Declarative
-import           Test.Hspec.Core.Spec (SpecM, SpecWith, runIO)
+import           Test.Hspec.Core.Spec (SpecM, SpecWith, runIO, modifyConfig)
 import qualified Test.Hspec.Core.Spec as Core
 import qualified Test.Hspec.Core.Config.Definition as Core
 
@@ -97,11 +97,7 @@ registerOptions = Core.modifyConfig . Core.addExtensionOptions section . map lif
     package = maybe "main" (CallStack.srcLocPackage . snd) CallStack.callSite
 
     liftOption :: Option -> Declarative.Option Core.Config
-    liftOption = Declarative.mapOption Config.from Config.to . Config.unOption
-
-
-modifyConfig :: (Config -> Config) -> SpecWith a
-modifyConfig f = Core.modifyConfig (Config.to . f . Config.from)
+    liftOption = Config.unOption
 
 {- |
 Register a transformation that transforms the spec tree in phase 3.
--- a/hspec-core/src/Test/Hspec/Core/Extension.hs
+++ b/hspec-core/src/Test/Hspec/Core/Extension.hs
@@ -1,5 +1,6 @@
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
 -- | Stability: unstable
-module Test.Hspec.Core.Extension (
+module Test.Hspec.Core.Extension {-# WARNING "This API is experimental." #-} (
 -- * Lifecycle of a test run
 {- |
 A test run goes through four distinct phases:
--- a/hspec-core/src/Test/Hspec/Core/Extension.hs
+++ b/hspec-core/src/Test/Hspec/Core/Extension.hs
@@ -16,6 +16,7 @@ When writing extensions the following imports are recommended:
 @
 import "Test.Hspec.Core.Extension"
 import "Test.Hspec.Core.Extension.Config" qualified as Config
+import "Test.Hspec.Core.Extension.Option" qualified as Option
 import "Test.Hspec.Core.Extension.Item" qualified as Item
 import "Test.Hspec.Core.Extension.Spec" qualified as Spec
 import "Test.Hspec.Core.Extension.Tree" qualified as Tree
@@ -37,16 +38,12 @@ import "Test.Hspec.Core.Extension.Tree" qualified as Tree
 
 -- ** Phase 2: Parsing command-line options
 {- |
-An extension can use `registerOption` during phase 1 to register custom command-line options, and as a consequence indirectly influence this phase.
+An extension can use `registerOptions` during phase 1 to register custom command-line options, and as a consequence indirectly influence this phase.
 
 * Options can use @Config.`Test.Hspec.Core.Extension.Config.setAnnotation`@ to add custom metadata to the `Config`.
 -}
 , Option
-, OptionSetter
-, registerOption
-, flag
-, option
-, argument
+, registerOptions
 
 -- ** Phase 3: Transforming the spec tree
 {- |
@@ -88,29 +85,19 @@ import qualified Test.Hspec.Core.Config.Definition as Core
 import           Test.Hspec.Core.Extension.Config (Config)
 import qualified Test.Hspec.Core.Extension.Config as Config
 import qualified Test.Hspec.Core.Extension.Config.Type as Config
+import           Test.Hspec.Core.Extension.Option
 import           Test.Hspec.Core.Extension.Item
 import           Test.Hspec.Core.Extension.Tree
 
-newtype Option = Option { unOption :: Declarative.Option Config }
-newtype OptionSetter = OptionSetter { unOptionSetter :: Declarative.OptionSetter Config }
-
-flag :: String -> (Bool -> Config -> Config) -> String -> Option
-flag name setter = Option . Core.flag name setter
-
-option :: String -> OptionSetter -> String -> Option
-option name setter = Option . Core.option name (unOptionSetter setter)
-
-argument :: String -> (String -> Maybe a) -> (a -> Config -> Config) -> OptionSetter
-argument name parser setter = OptionSetter (Core.argument name parser setter)
-
-registerOption :: HasCallStack => Option -> SpecWith a
-registerOption = Core.modifyConfig . Core.addExtensionOptions section . return . liftOption
+registerOptions :: HasCallStack => [Option] -> SpecWith a
+registerOptions = Core.modifyConfig . Core.addExtensionOptions section . map liftOption
   where
     section = "OPTIONS FOR " <> package
     package = maybe "main" (CallStack.srcLocPackage . snd) CallStack.callSite
 
     liftOption :: Option -> Declarative.Option Core.Config
-    liftOption = Declarative.mapOption Config.from Config.to . unOption
+    liftOption = Declarative.mapOption Config.from Config.to . Config.unOption
+
 
 modifyConfig :: (Config -> Config) -> SpecWith a
 modifyConfig f = Core.modifyConfig (Config.to . f . Config.from)
--- a/hspec-core/src/Test/Hspec/Core/Extension/Config/Type.hs
+++ b/hspec-core/src/Test/Hspec/Core/Extension/Config/Type.hs
@@ -1,11 +1,7 @@
-{-# LANGUAGE ViewPatterns #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# OPTIONS_GHC -fno-warn-deprecations #-}
 module Test.Hspec.Core.Extension.Config.Type (
   Option(..)
 , Config(..)
-, to
-, from
 
 , setAnnotation
 , getAnnotation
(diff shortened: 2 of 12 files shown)


BebeSparkelSparkel:
I think it would be helpful to have a specific environment lookup function or even better to have `flag`, `option`, ... automatically do an environment lookup. Either should have the standard HSPEC_ prefix.

sol:
> I think it would be helpful to have a specific environment lookup function or even better to have `flag`, `option`, ... automatically do an environment lookup. Either should have the standard HSPEC_ prefix.

I think that should already be the case, according to the rules documented here: https://hspec.github.io/options.html#specifying-options-through-environment-variables

sol:
The reason that the example  specifically looks up the `CI` environment variable is because this environment variable is already set when for  GitHub actions, but setting `HSPEC_CI=yes` will work as well.

BebeSparkelSparkel:
It would be nice if the `Test.Hspec.Core.Extension.option` interface did not take the type `OptionSetter` and instead have interface functions for the common cases.

sol:
> It would be nice if the `Test.Hspec.Core.Extension.option` interface did not take the type `OptionSetter` and instead have interface functions for the common cases.

How exactly would that look like?

Right now you can say things like

```haskell
registerOption $ flag "ci" setCiFlag
    "include itmes that are marked as \"CI only\""
```
or
```haskell
registerOption $ option "tags" (argument "TAGS" return addTagFilters)
    "TAGS can be a list of tag names."
```

How would you want to express this instead?

BebeSparkelSparkel:
In my pr I used the below

```
  , option "timeout" (argument "N" (fmap Seconds . readMaybe) (setTimeout . Just)) "each test will run at most N seconds"
   , mkOptionNoArg "no-timeout" Nothing (setTimeout Nothing) "remove test time limits"
```

So, it would be useful to have

```haskell
optionalArg :: Read a => String -> String -> (a -> Config -> Config) -> Option
optionNoArg :: String -> (Config -> Config) -> Option
```

these are more clear than trying to figure out which constructor to use with `option`. 

`option` and `argument`  could then be moved to `GetOpt.Declarative.Types` for more advanced uses.

BebeSparkelSparkel:
Another feature of extensions is that QuickCheck does not need to be a dependency of hspec-core anymore.

sol:
> ```
> mkOptionNoArg "no-timeout" Nothing (setTimeout Nothing) "remove test time limits"
> ```

I think this is currently not covered by the API, I'll try to look into it tomorrow.

hspec-core/src/Test/Hspec/Core/Extension.hs
```diff
@@ -0,0 +1,71 @@
...
+import           Test.Hspec.Core.Spec (SpecM, SpecWith, runIO)
+import qualified Test.Hspec.Core.Spec as Core
+import qualified Test.Hspec.Core.Config.Definition as Core
+import           Test.Hspec.Core.Extension.Config (Config)
+import qualified Test.Hspec.Core.Extension.Config as Config
+import qualified Test.Hspec.Core.Extension.Config.Type as Config
+import           Test.Hspec.Core.Extension.Item (Item)
+import           Test.Hspec.Core.Extension.Tree (SpecTree)
+
+newtype Option = Option { unOption :: Declarative.Option Config }
```
BebeSparkelSparkel, line 46: You have several well organized sub extension modules like `Item`, `Spec`, `Tree`, and `Config`. The option types and functions would be better organized in their own module and then just have this a reexport module for all the sub export modules and the required types/functions.

hspec-core/src/Test/Hspec/Core/Extension/Item.hs
```diff
@@ -0,0 +1,46 @@
...
+, getAnnotation
+) where
+
+import           Prelude ()
+import           Test.Hspec.Core.Compat
+
+import           Test.Hspec.Core.Spec hiding (pending, pendingWith)
+import           Test.Hspec.Core.Tree
+
+isFocused :: Item a -> Bool
```
BebeSparkelSparkel, line 27: I think extensions should be able to modify every field of `Item`.

hspec-core/src/Test/Hspec/Core/Extension/Spec.hs
```diff
@@ -0,0 +1,10 @@
+module Test.Hspec.Core.Extension.Spec (
+  mapItems
+) where
+
+import           Prelude ()
+
+import           Test.Hspec.Core.Spec
+
+mapItems :: (Item a -> Item b) -> SpecWith a -> SpecWith b
```
BebeSparkelSparkel, line 9: Seems like a better home for this is in hspec-core/src/Test/Hspec/Core/Extension/Item.hs

hspec-core/src/Test/Hspec/Core/Extension/Tree.hs
```diff
@@ -0,0 +1,19 @@
...
+import           Prelude ()
+import           Test.Hspec.Core.Compat
+
+import           Test.Hspec.Core.Spec hiding (SpecTree)
+import qualified Test.Hspec.Core.Spec as Core
+
+type SpecTree = Core.SpecTree ()
+
+mapItems :: (Item () -> Item ()) -> [SpecTree] -> [SpecTree]
+mapItems = bimapForest id
```
BebeSparkelSparkel, line 16: Wouldn't `fmap . fmap` be better?

BebeSparkelSparkel (changes requested):
Just some thoughts

hspec-core/src/Test/Hspec/Core/Config/Definition.hs
```diff
@@ -91,13 +126,11 @@ data Config = Config {
 , configFormatter :: Maybe V1.Formatter
 , configHtmlOutput :: Bool
 , configConcurrentJobs :: Maybe Int
+, configAnnotations :: Annotations
```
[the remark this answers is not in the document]
sol: Yes, exactly, however, this will need some more work: https://github.com/hspec/hspec/pull/870

hspec-core/src/Test/Hspec/Core/Extension/Item.hs
sol: I think they can do that already.  The `Item` type is exported.

hspec-core/src/Test/Hspec/Core/Extension/Spec.hs
sol: We have `Spec.mapItems` and `Tree.mapItems`, that's why they are in their respective modules.  I added some rudimentary documentation to hopefully outline this: https://github.com/hspec/hspec/pull/868/files

hspec-core/src/Test/Hspec/Core/Extension/Tree.hs
sol: I'm somewhat undecided, yes, you are right that's true, but would this make the code any more clean?

hspec-core/src/Test/Hspec/Core/Extension.hs
sol: I haven't looked into the options yet, I try to get to it sometime soon.

hspec-core/src/Test/Hspec/Core/Extension/Tree.hs
BebeSparkelSparkel: I think most people know what `fmap . fmap` does especially for the type `[SpecTree]`, but `bimapForest` is a specialized function that would require reference.

hspec-core/src/Test/Hspec/Core/Extension/Spec.hs
BebeSparkelSparkel: good point

hspec-core/src/Test/Hspec/Core/Extension/Item.hs
BebeSparkelSparkel: but the fields are not

sol: They are exported from `Test.Hspec.Core.Extension.Item`: https://github.com/hspec/hspec/pull/894/files#diff-edd9033d1fe072b0c2f04c047d3566881bbf49c63d961497a189d89854b9774bR3

hspec-core/src/Test/Hspec/Core/Extension/Tree.hs
sol: I guess `map . fmap` works for me.

hspec-core/src/Test/Hspec/Core/Extension/Config/Type.hs
```diff
@@ -0,0 +1,184 @@
...
+
+import           Test.Hspec.Core.Format
+import           Test.Hspec.Core.Config.Definition (ColorMode(..), UnicodeMode(..))
+import qualified Test.Hspec.Core.Config.Definition as Core
+import           Test.Hspec.Core.Annotations (Annotations)
+import qualified Test.Hspec.Core.Annotations as Annotations
+
+import           Test.Hspec.Core.Extension.Tree (SpecTree)
+
+data Config = Config {
```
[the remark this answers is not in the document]
sol: I think it's possible to remove some of those options after #870.  However, only from this `Config` type. The fields of the core `Config` can only be deprecated, not removed, for backwards compatibility.

That's also the reason why I duplicated the type, so that the extension API doesn't "know  about" legacy stuff.

It probably makes sense to finish #870 before the extensions API lands  on `main`.  

That's the approach I'm taking for now:
- The  provisional API is on #894, I use this as a staging ground
- #870 will go on #894 once it's done
- Any additional changes will be merged to #894
- Once everything looks good #894 will go on `main`


This PR is now against #894, with the options and documentation changes.

hspec-core/src/Test/Hspec/Core/Extension.hs
```diff
@@ -0,0 +1,71 @@
...
+newtype OptionSetter = OptionSetter { unOptionSetter :: Declarative.OptionSetter Config }
+
+flag :: String -> (Bool -> Config -> Config) -> String -> Option
+flag name setter = Option . Core.flag name setter
+
+option :: String -> OptionSetter -> String -> Option
```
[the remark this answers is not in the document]
sol: https://github.com/hspec/hspec/pull/868/files#diff-89c0bd2feff5876f008665eb349641509884e5274c5eac99342be545d21d4bc3

Is this along the lines of what you had in mind?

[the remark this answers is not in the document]
sol: I'm not 100% solved on it, as it's not as clear anymore, which string is the option name, and which string is the name for the parameter.  But let's give this  a try, provisionally.

hspec-core/src/Test/Hspec/Core/Extension.hs
```diff
@@ -0,0 +1,71 @@
...
+option name setter = Option . Core.option name (unOptionSetter setter)
+
+argument :: String -> (String -> Maybe a) -> (a -> Config -> Config) -> OptionSetter
+argument name parser setter = OptionSetter (Core.argument name parser setter)
+
+registerOption :: HasCallStack => Option -> SpecWith a
```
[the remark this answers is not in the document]
sol: It's now `registerOptions` that expects a list of `Option` values.

Keeping the `Option` type Might allow for reuse of `Test.Hspec.Core.Extension.Option` for https://github.com/hspec/hspec/pull/870.  So let's not abandon it, at least not yet.

hspec-core/src/Test/Hspec/Core/Extension/Config/Type.hs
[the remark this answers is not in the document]
BebeSparkelSparkel: Thank your for outlining the dev plan for extensions. Is there anything you want me to specifcally look at?

hspec-core/src/Test/Hspec/Core/Extension/Option.hs
```diff
@@ -0,0 +1,24 @@
...
+
+import qualified GetOpt.Declarative as Declarative
+import qualified Test.Hspec.Core.Config.Definition as Core
+
+import           Test.Hspec.Core.Extension.Config.Type (Option(..), Config)
+
+flag :: String -> (Bool -> Config -> Config) -> String -> Option
+flag name setter = Option . Core.flag name setter
+
+argument :: String -> String -> (String -> Config -> Maybe Config) -> String -> Option
```
BebeSparkelSparkel, line 17: You are right about the which string is which. Perhaps an aliases would be helpful.
```haskell
argument :: CliName -> CliArgName -> CliHelp -> (Bool -> Config -> Config) -> SpecWith a
type CliName = String
type CliArgName = String
type CliHelp = String
```

I rearranged the arguments because I like having functions at the end which allows multiline lambdas without parenthaseses and hanging arguments. This should be applied to all the option functions if you choose to use it.

I changed `Option` to `SpecWith a` which removes the need for `registerOption[s]` and allows these to be placed as their own line in `do`. This is not a big deal but does seem to be cleaner and less overhead for extension writers. Applies to all option functions.

hspec-core/src/Test/Hspec/Core/Extension.hs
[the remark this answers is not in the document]
BebeSparkelSparkel: You are right about the which string is which. Perhaps aliases would be helpful.
```haskell
argument :: CliName -> CliArgName -> CliHelp -> (Bool -> Config -> Config) -> Option
type CliName = String
type CliArgName = String
type CliHelp = String
```

I rearranged the arguments because I like having functions at the end which allows multiline lambdas without parenthaseses and hanging arguments. This should be applied to all the option functions if you choose to use it.

hspec-core/src/Test/Hspec/Core/Extension.hs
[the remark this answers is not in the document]
BebeSparkelSparkel: Ok sounds good.

sol:
@BebeSparkelSparkel OK, so I decided to:

- go without `OptionSetter` per your suggestion
- remove the second `Config`  type, to allow for a use case that came up in https://github.com/hspec/hspec/pull/895 (specifically `--add-format`)
- add WARNING pragmas to the extension modules, that clearly marke them as experimental.  This means that you have to add `{-# OPTIONS_GHC -fno-warn-deprecations #-}` to extensions until we fully fledged out the API; but it also means that we can release the extensions API early without overthinking stuff.


What I didn't do yet:
- add types synonyms for string option arguments
- change the order of option arguments (I am open to reconsider this later)
- change the type of option primitives from `Option` to `Spec` and get rid of `registerOption` (I am inclined to look at this again after #870)

I'm inclined to move forward with a release, so that we can gather some practical experience with the API.  Given that the API is guarded by  WARNING pragmas we are not committing to anything and can always change things.

parsonsmatt:
This is brilliant!! Brb gotta rewrite my flaky test detection and timeout logic for this...

sol:
@parsonsmatt feedback very much welcome please!

sol:
@parsonsmatt also look at https://github.com/hspec/hspec/pull/893/files for more examples.

parsonsmatt:
Yup! I'm going to rewrite my flaky test detection and our app's test timeout using this.

PR #877, opened 2024-02-24
Add do in 2nd test for consistency

OlegAlexander:
Hello and thank you for creating hspec!

As a new user of hspec, I noticed that the second example test doesn't have a `do` at the end. I know the `do` isn't necessary there, but it's also not necessary at the end of any of the tests. Therefore, I propose to make things consistent by adding a `do` at the end of the second test. 

Thank you.

### DIFF
--- a/doc/_includes/Example.hs
+++ b/doc/_includes/Example.hs
@@ -8,7 +8,7 @@ main = hspec $ do
     it "returns the first element of a list" $ do
       head [23 ..] `shouldBe` (23 :: Int)
 
-    it "returns the first element of an *arbitrary* list" $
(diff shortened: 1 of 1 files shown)


sol:
@OlegAlexander thanks for using `hspec`.

This will not fly:

1. The first example has type `IO ()`.
2. The type of the second example is `Property`.
3. There is a`Monad` instance for `IO`, but not for `Property`.

So while it's synthetically valid to put the `do` for the second example, it's never useful. For the first and third example however, you may want to add additional statements to the `do`-block at a later point. Here always adding a `do` will give you smaller diffs down the road.

OlegAlexander:
Ok. Thank you for the explanation!

PR #881, opened 2024-04-20
Implement simple SpecWrap

### LINKED ISSUE #880: Spec hook for each discovered file in `hspec-discover`
I'm experimenting with reducing module-level dependencies in our test suite.

I've got something I'm pretty happy with in the `SpecHook` facility. However, it requires that a directory-at-a-time is converted, which is a pretty big chunk of work all at once. 

I can write a polymorphic hook function `provide :: (Provide arg) => SpecWith arg -> Spec`. This lets users define their `spec :: SpecWith Foo` as long as there's an `instance Provide Foo` somewhere in `Spec`. However, I can't use this with `SpecHook`.

Looking at the code, the "issue" is that each `SpecHook` creates a node in a tree - so you'd get:

```haskell
spec :: Spec
spec = do
  Foo.SpecHook.hook do
    Foo.BarSpec.spec
    Foo.BazSpec.spec
  BlahSpec.spec
```

If `Foo.BarSpec` and `Foo.BazSpec` are typed differently, then this fails.

What I want instead is to be able to have a `EachSpecHook` that is applied to every single test individually. The above example would then look like:

```haskell
spec :: Spec
spec = do
  Foo.SpecEachHook.hook Foo.BarSpec.spec
  Foo.SpecEachHook.hook Foo.BazSpec.spec
  BlahSpec.spec
```

This would allow `hook` to be polymorphic and work even when `Foo.BarSpec` or `Foo.BazSpec` differ on type.

parsonsmatt:
This PR fixes #880 in a very simple way. If a module `SpecWrap` is present next to the `hspec-discover` module, then each `Spec` will be wrapped in the function `SpecWrap.wrapSpec`.


sol:
@parsonsmatt I'll try to look at this, hopefully on Monday, sorry for the delay.

sol:
@parsonsmatt from what I understand, unlike `SpecHook.hs`, `SpecWrap.hs` is only discovered at the top-level.  Is this in an attempt to keep things simple, or is there a more fundamental reason for this?

As for the change in general, I'm eager to keep the number of concepts that `hspec-discover` supports within reasonable limits.  For that reason:

1. I prefer calling things more in line with  what you originally suggested in #880. Specifically I prefer `SpecHookEach.hs` over `SpecWrap.hs`. It should then behave exactly  the same as `SpecHook.hs`, with the only difference being that it is applied to each spec module individually.
2. An alternative approach that I'm toying around with mentally is to refactor `hspec-discover` in a way that makes it easy to create custom "discover"-executables.

(2) could be something along the lines of:
```haskell
runWith :: Config -> IO ()
runWith = ...

data Config = Config {
  discoverModules :: ModuleName -> Bool
, discoverHooks :: ModuleName -> Bool
, renderAdditionalImports :: [ModuleName]
, renderCombine :: Combine
, renderDriverFunction :: Identifier
, renderLabelWith :: Identifier
, renderModuleExport :: Identifier
, renderApplyHooks :: ApplyHooks
, renderHookModuleExport :: Identifier
}

data Combine = Monadic | AsList

data ApplyHooks = Once | AroundEach

defaultConfig :: Config
defaultConfig = Config {
  discoverModules = endsWith "Spec"
, discoverHooks = (== "SpecHook")
, renderAdditionalImports = ["Test.Hspec"]
, renderCombine = Monadic
, renderDriverFunction = "hspec"
, renderLabelWith = "describe"
, renderModuleExport = "spec"
, renderApplyHooks = Once
, renderHookModuleExport = "hook"
}
```

Where your specific use case would boil down to:

```haskell
main :: IO ()
main = runWith config

config :: Config
config = defaultConfig {
  discoverHooks = (== "SpecWrap")
, renderApplyHooks = AroundEach
, renderHookModuleExport = "wrapSpec"
}
```

PR #890, opened 2024-05-22
added test run time limits

### LINKED ISSUE #889: Limit Duration of Each Individual Test
It would be useful to be able to limit each test to a number of seconds.

BebeSparkelSparkel:
Allow tests to have limited run time via a command line argument or Spec modifiers.

closes #889

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Clock.hs
+++ b/hspec-core/src/Test/Hspec/Core/Clock.hs
@@ -6,6 +6,7 @@ module Test.Hspec.Core.Clock (
 , toMicroseconds
 , getMonotonicTime
 , measure
+, measureWithTimeout
 , sleep
 , timeout
 ) where
@@ -42,9 +43,12 @@ getMonotonicTime = do
 #endif
 
 measure :: IO a -> IO (Seconds, a)
-measure action = do
+measure = fmap (second fromJust) . measureWithTimeout Nothing
+
+measureWithTimeout :: Maybe Seconds -> IO a -> IO (Seconds, Maybe a)
+measureWithTimeout maxTime action = do
   t0 <- getMonotonicTime
-  a <- action
+  a <- maybe (fmap Just) timeout maxTime action
   t1 <- getMonotonicTime
   return (t1 - t0, a)
 
--- a/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
+++ b/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
@@ -26,7 +26,7 @@ import           System.Directory (getTemporaryDirectory, removeFile)
 import           System.IO (openTempFile, hClose)
 import           System.Process (system)
 
-import           Test.Hspec.Core.Format (Format, FormatConfig)
+import           Test.Hspec.Core.Format (Format, FormatConfig, Seconds(Seconds))
 import           Test.Hspec.Core.Formatters.Pretty (pretty2)
 import qualified Test.Hspec.Core.Formatters.V1.Monad as V1
 import           Test.Hspec.Core.Util
@@ -48,6 +48,7 @@ data Config = Config {
 , configFailOnFocused :: Bool
 , configFailOnPending :: Bool
 , configFailOnEmptyDescription :: Bool
+, configTimeout :: Maybe Seconds
 , configPrintSlowItems :: Maybe Int
 , configPrintCpuTime :: Bool
 , configFailFast :: Bool
@@ -107,6 +108,7 @@ mkDefaultConfig formatters = Config {
 , configFailOnFocused = False
 , configFailOnPending = False
 , configFailOnEmptyDescription = False
+, configTimeout = Nothing
 , configPrintSlowItems = Nothing
 , configPrintCpuTime = False
 , configFailFast = False
@@ -304,6 +306,9 @@ setMaxShrinks n c = c {configQuickCheckMaxShrinks = Just n}
 setSeed :: Integer -> Config -> Config
 setSeed n c = c {configSeed = Just n}
 
+setTimeout :: Maybe Seconds -> Config -> Config
+setTimeout n c = c {configTimeout = n}
+
 data FailOn =
     FailOnEmpty
   | FailOnFocused
@@ -345,6 +350,9 @@ runnerOptions = [
   , mkOption    "fail-on" Nothing (argument "ITEMS" readFailOnItems (setFailOnItems True )) helpForFailOn
   , mkOption "no-fail-on" Nothing (argument "ITEMS" readFailOnItems (setFailOnItems False)) helpForFailOn
   , flag "strict" setStrict $ "same as --fail-on=" <> showFailOnItems strict
+
+  , option "timeout" (argument "N" (fmap Seconds . readMaybe) (setTimeout . Just)) "each test will run at most N seconds"
+  , mkOptionNoArg "no-timeout" Nothing (setTimeout Nothing) "remove test time limits"
 
   , flag "fail-fast" setFailFast "abort on first failure"
   , flag "randomize" setRandomize "randomize execution order"
--- a/hspec-core/src/Test/Hspec/Core/Runner.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner.hs
@@ -115,7 +115,7 @@ import qualified Test.QuickCheck as QC
 import           Test.Hspec.Core.Util (Path)
 import           Test.Hspec.Core.Clock
 import           Test.Hspec.Core.Spec hiding (pruneTree, pruneForest)
-import           Test.Hspec.Core.Tree (formatDefaultDescription)
+import           Test.Hspec.Core.Tree (formatDefaultDescription, setItemTimeout)
 import           Test.Hspec.Core.Config
 import           Test.Hspec.Core.Format (Format, FormatConfig(..))
 import qualified Test.Hspec.Core.Formatters.V1 as V1
@@ -233,11 +233,14 @@ hspecWithResult config = fmap toSummary . hspecWithSpecResult config
 
 hspecWithSpecResult :: Config -> Spec -> IO SpecResult
 hspecWithSpecResult defaults spec = do
-  (c, forest) <- evalSpec defaults spec
+  (c, f) <- evalSpec defaults spec
   config <- getArgs >>= readConfig c
   oldFailureReport <- readFailureReportOnRerun config
 
   let
+    forest :: [SpecTree ()]
+    forest = maybe id (fmap . fmap . setItemTimeout . Just) (configTimeout config) f
+
     normalMode :: IO SpecResult
     normalMode = doNotLeakCommandLineArgumentsToExamples $ runSpecForest_ oldFailureReport forest config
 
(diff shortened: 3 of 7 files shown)


hspec-core/src/Test/Hspec/Core/Config/Definition.hs
```diff
@@ -353,6 +358,7 @@ runnerOptions = [
   , mkOptionNoArg "rerun-all-on-success" Nothing setRerunAllOnSuccess "run the whole test suite after a previously failing rerun succeeds for the first time (only works in combination with --rerun)"
   , mkOption "jobs" (Just 'j') (argument "N" readMaxJobs setMaxJobs) "run at most N parallelizable tests simultaneously (default: number of available processors)"
   , option "seed" (argument "N" readMaybe setSeed) "used seed for --randomize and QuickCheck properties"
+  , option "max-time-per-test" (argument "N" (fmap Seconds . readMaybe) setMaxTimePerTest) "each test will run at most N seconds"
```
sol, line 361: What do you think about naming the option `--timeout` instead of `--max-time-per-test`? Would that work for you? 

I think we also want `--no-timeout` to set it back to `Nothing`. Config options [can come from various places](https://hspec.github.io/options.html#precedence-resolution-for-options-that-are-specified-in-multiple-places), so you may want to set it back to nothing at the command line.

sol (changes requested):
This looks great!

- [ ] Please split this into two separate PRs, that is move everything "per item" into a new "draft" PR (or keep them locally for now, whatever works best for you).  Let's then get the command line option merged first and do "per item" as a follow up.
- [ ] This needs a tests! Maybe just a single acceptance test will be fine, look at https://github.com/hspec/hspec/blob/main/hspec-core/test/Test/Hspec/Core/RunnerSpec.hs

NB: Things like timeouts are tricky  to test, as we don't want to unnecessarily increase the runtime of the test suit. Still, maybe a timeout value of zero will do the trick, or maybe we will remove the test again later.

BebeSparkelSparkel:
I added tests but am using `hspecWithResult` which mucks up the test results. Is there a function that does not produce output or a function that suppresses the output?

sol:
> I am unsure how my changes caused _api-check / api-check (pull_request)_ to fail

This is due to GHC 9.10.1.  I fixed that with #892, please rebase once this is on `main`.

I also added a section to the README on how to update API dumps.  You will need to run that. The purpose of keeping track of API changes explicitly is so that we don't  expose  stuff that we didn't meant to, and to make it easier to determine whether we need a major or minor version bump.

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -450,11 +453,11 @@ toEvalItemForest :: Params -> [SpecTree ()] -> [EvalItemTree]
 toEvalItemForest params = bimapForest id toEvalItem . filterForest itemIsFocused
   where
     toEvalItem :: Item () -> EvalItem
-    toEvalItem (Item requirement loc isParallelizable _isFocused e) = EvalItem {
+    toEvalItem (Item requirement loc isParallelizable _isFocused e maxTime) = EvalItem {
       evalItemDescription = requirement
     , evalItemLocation = loc
     , evalItemConcurrency = if isParallelizable == Just True then Concurrent else Sequential
-    , evalItemAction = \ progress -> measure $ e params withUnit progress
+    , evalItemAction = \ progress -> fmap (fmap $ fromMaybe $ Result requirement $ Failure loc $ Reason "exceeded timeout") $ measureWithTimeout maxTime $ e params withUnit progress
```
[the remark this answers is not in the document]
sol: I think that's due to GHC 9.10.1, should be fixed now.

hspec-core/src/Test/Hspec/Core/Clock.hs
```diff
@@ -48,6 +49,13 @@ measure action = do
   t1 <- getMonotonicTime
   return (t1 - t0, a)
 
+measureWithTimeout :: Maybe Seconds -> IO a -> IO (Seconds, Maybe a)
+measureWithTimeout maxTime action = do
+  t0 <- getMonotonicTime
+  a <- maybe (fmap Just) timeout maxTime action
+  t1 <- getMonotonicTime
+  return (t1 - t0, a)
```
sol, line 57: Is it possible to implement this in terms of `measure`?

```haskell
measureWithTimeout maxTime action = measure $ maybe (fmap Just) timeout maxTime action
```

hspec-core/src/Test/Hspec/Core/Config/Definition.hs
```diff
@@ -91,6 +91,7 @@ data Config = Config {
 , configFormatter :: Maybe V1.Formatter
 , configHtmlOutput :: Bool
 , configConcurrentJobs :: Maybe Int
+, configMaxTimePerTest :: Maybe Seconds
```
sol, line 94: Please move this closer to related options, e.g. after `configFailOnEmptyDescription` (matching the location in the help message).

hspec-core/src/Test/Hspec/Core/Config/Definition.hs
```diff
@@ -138,6 +139,7 @@ mkDefaultConfig formatters = Config {
 , configFormatter = Nothing
 , configHtmlOutput = False
 , configConcurrentJobs = Nothing
+, configMaxTimePerTest = Nothing
```
sol, line 142: Same here

hspec-core/test/Test/Hspec/Core/RunnerSpec.hs
```diff
@@ -1021,3 +1021,15 @@ spec = do
     context "on failure" $ do
       it "returns False" $ do
         H.rerunAll config (Just report) result { specResultSuccess = False } `shouldBe` False
+
+  describe "timeout" $ do
```
sol, line 1025: Please move this test case closer to tests for related options, e.g. around line 389.

hspec-core/src/Test/Hspec/Core/Runner.hs
```diff
@@ -233,11 +233,14 @@ hspecWithResult config = fmap toSummary . hspecWithSpecResult config
...
 hspecWithSpecResult :: Config -> Spec -> IO SpecResult
 hspecWithSpecResult defaults spec = do
-  (c, forest) <- evalSpec defaults spec
+  (c, f) <- evalSpec defaults spec
   config <- getArgs >>= readConfig c
   oldFailureReport <- readFailureReportOnRerun config
 
   let
+    forest :: [SpecTree ()]
+    forest = maybe id (fmap . fmap . setItemMaxTime) (configMaxTimePerTest config) f
```
sol, line 242: I think this is not strictly needed and can be deferred to #891.

hspec-core/src/Test/Hspec/Core/Runner.hs
sol, line 460: As I understand it, you could use  `maxTime` instead of `configMaxTimePerTest config` here, for now.

Also, that line of code is getting quite long. I think it could make sense to extract some of the code into a separate function, say `applyTimeout` or `failOnTimeout` or both, not sure exactly how it will turn out nice, maybe just try something.

hspec-core/src/Test/Hspec/Core/Tree.hs
```diff
@@ -135,6 +143,7 @@ specItem s e = Leaf Item {
   , itemIsParallelizable = Nothing
   , itemIsFocused = False
   , itemExample = safeEvaluateExample e
+  , itemMaxTime = Nothing
```
sol, line 146: Same here, I think this can be deferred to #891.

hspec-core/src/Test/Hspec/Core/Config/Definition.hs
sol: I try to keep names of command line options and the corresponding fields of the config data type consistent.  So this should be `configTimeout`, or if this is too generic then alternatively `configSpecItemTimeout`. I would probably go with the first.

sol (changes requested):
Hey, this looks great. I left a couple of more comments. I haven't looked at a test case in detail yet, but from a quick glimpse, I assume this doesn't add any delays to the test suite, which is nice. Please  take a look at my comments and make sure CI is green, then I will take a final look.

BebeSparkelSparkel:
I'm not sure how to accommodate the unresolved. Also, if #891 is to be pulled then not storing the timeout in Item is extra work that will be undone.

sol:
@BebeSparkelSparkel you marked a couple of my comments as resolved, but I don't see any associated code changes. Did you forget to push?

BebeSparkelSparkel:
Changed the wrong branch. I'll try and git fix it.

hspec-core/src/Test/Hspec/Core/Runner.hs
BebeSparkelSparkel: > As I understand it, you could use  maxTime instead of configMaxTimePerTest config here, for now.

Sorry, I don't understand this.

hspec-core/src/Test/Hspec/Core/Tree.hs
BebeSparkelSparkel: I am unsure where to get the timeout duration from if this is removed.

sol:
I finalized an extension API that allows additional functionality to be implemented outside of `hspec-core`: https://github.com/hspec/hspec/pull/868

Originally, I didn't want to delay your feature until the API is finalized, but now that I consider it done I think the way forward is to implement this feature as a separate package. This also means that you have complete freedom on how to name things, coding style, etc + you are not blocked by waiting for a code review from me.

@BebeSparkelSparkel there are two options here, either you implement your feature with the git version of `hspec-core`.  This will give you the opportunity to  provide  feedback before the API is released. Or alternatively, I can move forward with a release and you can then use the version from Hackage.  Please let me know what you prefer.

Once your package is on Hackage, I will then add it to Hspec's documentation, so that it can be used by others.

sol:
@BebeSparkelSparkel the PR description contains an example on how to use the API + here is an other example: https://github.com/hspec/hspec/blob/ec2c16712ef68ba41d6f721d32840afadcba9576/hspec-core/src/Test/Hspec/Tags.hs

BebeSparkelSparkel:
I will try the git version implementation and add comment to the PR if I have issues.

Since you were considering adding this to core, should I make this a new package in this repo since this is a fundamental part of testing?

sol:
> I will try the git version implementation and add comment to the PR if I have issues.
> 
> Since you were considering adding this to core, should I make this a new package in this repo since this is a fundamental part of testing?

I suggest a separate package in a separate repo.  We can always decide to upstream things later.  I could imagine to upstream `--timeout` eventually, but maybe not per item timeouts, they could remain in a separate package and I imagine we could still make both of them  interact  nicely with each other. But let's save this discussion for another day.

By the way, if you think a separate repo is too much overhead, you could add it to `hspec-contrib`. I still think a separate package will  provide more visibility. `hspec-contrib` is just a dumping ground for random stuff.

BebeSparkelSparkel:
Ok, do extensions also allow for the implementation of `limit` and `noLimit` that can override the the global ``--timeout`?

sol:
> Ok, do extensions also allow for the implementation of `limit` and `noLimit` that can override the the global `--timeout`?

Yes, I think so. Look at the "tags" example.

sol:
> > Ok, do extensions also allow for the implementation of `limit` and `noLimit` that can override the the global `--timeout`?
> 
> Yes, I think so. Look at the "tags" example.

@BebeSparkelSparkel I think even the "ci" example should demonstrate all the ideas, but it's much shorter. So maybe that's a better place to start. Ask if you need any specific advice or help.

BebeSparkelSparkel:
Thank you I will attempt this now

PR #894, opened 2024-05-30
Make Hspec extensible

sol:
- [x] Allow for extensions that transform the spec tree
- [ ] Allow custom command-line options for `Example` instances


BebeSparkelSparkel:
It is unclear to me what this pr is trying to accomplish that is different from #868

PR #895, opened 2024-06-08
Support multiple --formats at once

pbrisbin:
### [Support passing --format more than once](https://github.com/hspec/hspec/pull/895/commits/5ee26a5f9d07c3b74fb9334fafdf4a004392e280)
[5ee26a5](https://github.com/hspec/hspec/pull/895/commits/5ee26a5f9d07c3b74fb9334fafdf4a004392e280)

We use a JUnit formatter that produces an XML file for external
analysis. When we do this, we'd like to still output the normal format
in the build logs, since viewing them there is also useful. Many of our
Engineers get confused when their build fails tests and they navigate to
the GitHub action and see no output. The summary and annotations
produced from our JUnit report are both less intuitive to find (although
more useful once they are found, which is why we do it).

This commit changes the behavior when passing `--format` multiple times
from "last wins" to "use them all". That way, we can pass `--format
checks --format junit` to achieve what we want.

### [Add --add-format option](https://github.com/hspec/hspec/pull/895/commits/57555adffa40c7f07d520d051c60da38f4b44ad1)
[57555ad](https://github.com/hspec/hspec/pull/895/commits/57555adffa40c7f07d520d051c60da38f4b44ad1)

This option can be used when you want to use a formatter in addition to
the default. Without this option, one would have to use `--format checks
--format junit` to get that behavior, which duplicates knowledge of the
default formatter and can easily be missed. Now, they can use
`--add-format junit` to accomplish it more directly.

The flag and help are placeholder; I'm happy to name it and document it
however you like.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
+++ b/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
@@ -177,7 +177,7 @@ argument name parser setter = Arg name $ \ input c -> flip setter c <$> parser i
 
 formatterOptions :: [(String, FormatConfig -> IO Format)] -> [Option Config]
 formatterOptions formatters = [
-    mkOption "format" (Just 'f') (argument "NAME" readFormatter setFormatter) helpForFormat
+    mkOption "format" (Just 'f') (argument "NAME" readFormatter addFormatter) helpForFormat
   , flag "color" setColor "colorize the output"
   , flag "unicode" setUnicode "output unicode"
   , flag "diff" setDiff "show colorized diffs"
@@ -215,8 +215,21 @@ formatterOptions formatters = [
     readFormatter :: String -> Maybe (FormatConfig -> IO Format)
     readFormatter = (`lookup` formatters)
 
-    setFormatter :: (FormatConfig -> IO Format) -> Config -> Config
-    setFormatter f c = c {configFormat = Just f}
+    addFormatter :: (FormatConfig -> IO Format) -> Config -> Config
+    addFormatter f config = config
+      { configFormat = Just $ maybe f (`combineFormats` f) $ configFormat config
+      }
+
+    combineFormats
+      :: (FormatConfig -> IO Format)
+      -> (FormatConfig -> IO Format)
+      -> FormatConfig -> IO Format
+    combineFormats f1 f2 config = do
+      formatEvent1 <- f1 config
+      formatEvent2 <- f2 config
+      pure $ \event -> do
+        formatEvent1 event
+        formatEvent2 event
 
     setColor :: Bool -> Config -> Config
     setColor v config = config {configColorMode = if v then ColorAlways else ColorNever}
--- a/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
+++ b/hspec-core/src/Test/Hspec/Core/Config/Definition.hs
@@ -29,6 +29,7 @@ import           System.Process (system)
 import           Test.Hspec.Core.Format (Format, FormatConfig)
 import           Test.Hspec.Core.Formatters.Pretty (pretty2)
 import qualified Test.Hspec.Core.Formatters.V1.Monad as V1
+import qualified Test.Hspec.Core.Formatters.V2 as V2
 import           Test.Hspec.Core.Util
 
 import           GetOpt.Declarative
@@ -178,6 +179,7 @@ argument name parser setter = Arg name $ \ input c -> flip setter c <$> parser i
 formatterOptions :: [(String, FormatConfig -> IO Format)] -> [Option Config]
 formatterOptions formatters = [
     mkOption "format" (Just 'f') (argument "NAME" readFormatter addFormatter) helpForFormat
+  , mkOption "add-format" Nothing (argument "NAME" readFormatter $ \f -> addFormatter f . defFormatter) helpForFormat
   , flag "color" setColor "colorize the output"
   , flag "unicode" setUnicode "output unicode"
   , flag "diff" setDiff "show colorized diffs"
@@ -214,6 +216,11 @@ formatterOptions formatters = [
 
     readFormatter :: String -> Maybe (FormatConfig -> IO Format)
     readFormatter = (`lookup` formatters)
+
+    defFormatter :: Config -> Config
+    defFormatter config = config
+      { configFormat = configFormat config <|> Just (V2.formatterToFormat V2.checks)
+      }
 
     addFormatter :: (FormatConfig -> IO Format) -> Config -> Config
     addFormatter f config = config


sol (commented):
@pbrisbin first things first, it's important to me that we end up with something that works flawlessly for you. So even I'm going to push back on this, please be very vocal if in the end your case isn't fully addressed.

From what I understand, this is specifically useful for [hspec-junit-formatter](https://hackage.haskell.org/package/hspec-junit-formatter), but not so much for other formatters. Do you think that's a fair assessment?

1. I'm not eager to change `--format`in a breaking way, and from what I understand, you wouldn't even need that for you use case. Is that correct?
2. While `--add-format` is technically still on the table, I'm eager to make it easier to extend Hspec with third-party packages.  Assuming that this is indeed mostly useful for [hspec-junit-formatter](https://hackage.haskell.org/package/hspec-junit-formatter), did you try to implement this in [hspec-junit-formatter](https://hackage.haskell.org/package/hspec-junit-formatter)? If yes, how  did it go? If no, can you give that a quick try and report back?

pbrisbin:
> I'm not eager to change `--format` in a breaking way, and from what I understand, you wouldn't even need that for you use case. Is that correct?

That's correct. Our current use case always wants to preserve the default formatter. I just found `--add-format` to be a bit odd, so wanted to offer only the `--format` change as an option to consider.

I totally get not wanting to break it, and that you consider this breaking. I think it's debatable, since passing it more than once is not something there is a real use case to do today, but I get it.

Would you accept changing `--format`'s usage from `NAME` to `NAME[,NAME...]`? Then we could use `--format checks,junit` to do what we want. This would be a bigger diff, a little more complexity, but I'm happy to write that code. And I don't think you'd consider this breaking.

> did you try to implement this in [hspec-junit-formatter](https://hackage.haskell.org/package/hspec-junit-formatter)? If yes, how did it go? If no, can you give that a quick try and report back?

If by "this" you mean "a custom formatter that outputs like `checks` and produces a file like `junit`", then I'm confident I could do that fine, I don't really need to try it, IMO. But I just don't think it makes architectural sense to do so:

1. Why not follow single-responsibility and have the "JUnit formatter" responsible for only JUnit formatting? What happens when `checks` evolves or is replaced as the default in hspec, now I have to react in junit-formatter?
2. I get that no formatters exist within hspec that make sense to use together, but surely it makes sense as a general use-case. I have my JUnit example here, but I could see a formatter that does profiling, or literally any other thing that's not related to actual output.

I would probably end up exposing a number of formatters under names that imply the various compositions: `junit` (JUnit report only), `checks+junit`, `progress+junit`, etc to cover all the bases. It just feels a bit janky to me, when support like I've added in this PR seems very low-effort.

> I'm eager to make it easier to extend Hspec with third-party packages

I wonder if it just feels a bit off because I'm using this idea of a "formatter" to run not-formatting things on Hspec events. The name implies it's about formatting test results for display to a human, in which case using more than one at once makes little sense. But I think think it may just be the naming that's wrong. What a "formatter" actually does is allow attaching any `Event -> IO ()` to the test run, which has myriad use-cases, IMO, and it would make total sense to compose many together.

pbrisbin:
> If by "this" you mean...

On second read, I think you were suggesting that hspec-junit-format expose a formatting module something like this:

```hs
module Test.Hspec.JUnit.Formatter
  ( use
  , add -- <-- diverge from the usual interface
  , register
  , format
  , module Api
  ) where

-- | Add JUnit formatting on top of the default formatting (checks)
add :: Spec -> Spec
add = undefined
```

Then clients could use `add` (instead of `use` or `register`) as their `SpecHook.hook`.

This seems fine enough; we could definitely do that. It would necessarily duplicate the knowledge that `checks` is the default formatter, since I'd have to write the `defFormatter` function of this PR there in order to achieve it, but that's OK.

The main limitation is you get `checks` (or whatever you give as `--format`) and `junit` all the time no matter what -- unless we implemented our own way to decide at run-time (e.g. ENV vars), which doesn't seem ideal.

So it just doesn't perfect solve wanting to do something like this:

- Run with `checks` in local development
- Run with `progress` and `junit` on CI
- Run with `checks` and `junit` in local development, sometimes

The intent behind the PR solves that perfectly, intuitively (IMO), and makes the JUnit formatter itself simpler.

pbrisbin:
OK, so I ran that down to get something working. It ends up looking like this in end-user code:

```hs
module SpecHook
  ( hook
  ) where

import Prelude

import Test.Hspec
import Test.Hspec.JUnit (defaultJUnitConfig)
import Test.Hspec.JUnit.Config.Env (envJUnitEnabled)
import qualified Test.Hspec.JUnit.Formatter as JUnit

hook :: Spec -> Spec
hook spec = do
  enabled <- runIO envJUnitEnabled
  if enabled
    then JUnit.add (defaultJUnitConfig "hspec") spec
    else spec
```

I'd rather not have to copy this from app to app where we want this functionality, but it does work.

In order to write `JUnit.add`, I did make some changes in hspec:

```hs
module Test.Hspec.JUnit.Formatter
  ( use
  , add
  , register
  , formatter
  , module Api
  ) where

import Prelude

import Test.Hspec.Api.Format.V2 as Api
import Test.Hspec.JUnit (JUnitConfig, junitFormat)

use :: JUnitConfig -> SpecWith a -> SpecWith a
use config = (modifyConfig (useFormatter $ formatter config) >>)

add :: JUnitConfig -> SpecWith a -> SpecWith a
add config = (modifyConfig (addFormatter $ formatter config) >>)

register :: JUnitConfig -> SpecWith a -> SpecWith a
register config = (modifyConfig (registerFormatter $ formatter config) >>)

formatter :: JUnitConfig -> (String, FormatConfig -> IO Format)
formatter config = ("junit", junitFormat config)
```

```diff
diff --git i/hspec-api/src/Test/Hspec/Api/Format/V1.hs w/hspec-api/src/Test/Hspec/Api/Format/V1.hs
index 74a6993e..a529f3d5 100644
--- i/hspec-api/src/Test/Hspec/Api/Format/V1.hs
+++ w/hspec-api/src/Test/Hspec/Api/Format/V1.hs
@@ -18,6 +18,7 @@ module Test.Hspec.Api.Format.V1 (
 -- * Register a formatter
 , registerFormatter
 , useFormatter
+, addFormatter
 , liftFormatter
 
 -- * Re-exports
@@ -31,6 +32,7 @@ import qualified Test.Hspec.Api.Format.V2 as V2
 import           Test.Hspec.Api.Format.V2 hiding (
     registerFormatter
   , useFormatter
+  , addFormatter
   , liftFormatter
   , FormatConfig(..)
   , Item(..)
@@ -54,6 +56,9 @@ registerFormatter = V2.registerFormatter . liftFormatterToV2
 useFormatter :: (String, FormatConfig -> IO Format) -> Config -> Config
 useFormatter = V2.useFormatter . liftFormatterToV2
 
+addFormatter :: (String, FormatConfig -> IO Format) -> Config -> Config
+addFormatter = V2.addFormatter . liftFormatterToV2
+
 -- | Make a formatter compatible with types from "Test.Hspec.Core.Format".
 liftFormatter :: (String, FormatConfig -> IO Format) -> (String, Latest.FormatConfig -> IO Latest.Format)
 liftFormatter = V2.liftFormatter . liftFormatterToV2
diff --git i/hspec-api/src/Test/Hspec/Api/Format/V2.hs w/hspec-api/src/Test/Hspec/Api/Format/V2.hs
index 1659cfad..ccf9d1fd 100644
--- i/hspec-api/src/Test/Hspec/Api/Format/V2.hs
+++ w/hspec-api/src/Test/Hspec/Api/Format/V2.hs
@@ -18,6 +18,7 @@ module Test.Hspec.Api.Format.V2 (
 -- * Register a formatter
 , registerFormatter
 , useFormatter
+, addFormatter
 , liftFormatter
 
 -- * Re-exports
@@ -43,6 +44,24 @@ registerFormatter = registerFormatter_ . liftFormatter
 useFormatter :: (String, FormatConfig -> IO Format) -> Config -> Config
 useFormatter (liftFormatter -> formatter@(_, format)) config = (registerFormatter_ formatter config) { configFormat = Just format }
 
+addFormatter :: (String, FormatConfig -> IO Format) -> Config -> Config
+addFormatter (liftFormatter -> formatter@(_, format)) config = (registerFormatter_ formatter config) { configFormat = configFormat' }
+ where
+  configFormat' = Just
+    $ maybe format (`addFormat` format)
+    $ configFormat config
+
+  addFormat
+    :: (Latest.FormatConfig -> IO Format)
+    -> (Latest.FormatConfig -> IO Format)
+    -> Latest.FormatConfig -> IO Format
+  addFormat f1 f2 fc = do
+    formatEvent1 <- f1 fc
+    formatEvent2 <- f2 fc
+    pure $ \event -> do
+      formatEvent1 event
+      formatEvent2 event
+
 -- copy of Test.Hspec.Core.Runner.registerFormatter
 registerFormatter_ :: (String, Latest.FormatConfig -> IO Latest.Format) -> Config -> Config
 registerFormatter_ formatter config = config { configAvailableFormatters = formatter : configAvailableFormatters config }
```

I still prefer not doing this, and instead having something like `--format=checks,junit` available, but here's what it looked like.

But I could also write that code directly in hspec-junit-formatter if you didn't want to accept it.

pbrisbin:
OK, sorry for all the comments, but I continued to move forward on this and I think we can close this PR and I'll supply the things I need in `hspec-junit-formatter` itself.

Here is my [current draft](https://github.com/freckle/hspec-junit-formatter/pull/28/files#diff-b6c7e05982d77f31be186a178c8c650433fe73d873cec65a3ec204c8cd3e9607). As noted in the docs, it supports three use-cases:

- Only produce a JUnit report (`use`)
- Produce a JUnit report in addition to normal output (`add`)
- Only produce a JUnit report, only if `--format=junit` is used (`register`)

It does not satisfy the following use-case:

- Produce a JUnit report in addition to normal output, only if `--format={something}` is used

That would be the only use-case that I think a PR like this would be required for, and I don't think it's an important one.

sol:
> I'll supply the things I need in `hspec-junit-formatter` itself.

I think that's the right call for now. We can always upstream `addFormatter` if we have a second use-case for it.

> It does not satisfy the following use-case:
> 
> * Produce a JUnit report in addition to normal output, only if `--format={something}` is used

I see a chance that this can be solved once I am done with #894. I think what you basically would need is a version of `modifyConfig` that is applied "late".

sol:
> > I'll supply the things I need in `hspec-junit-formatter` itself.
> 
> I think that's the right call for now. We can always upstream `addFormatter` if we have a second use-case for it.
> 
> > It does not satisfy the following use-case:
> > 
> > * Produce a JUnit report in addition to normal output, only if `--format={something}` is used
> 
> I see a chance that this can be solved once I am done with #894. I think what you basically would need is a version of `modifyConfig` that is applied "late".

Thinking about it, in retrospect `modifyConfig` should really be called `modifyDefaultConfig`.

PR #912, opened 2025-03-15
Add --no-rerun and --no-rerun-all-on-success (fixes #911)

### LINKED ISSUE #911: Implement success rerun `--no-rerun-all-on-success`
Is there a way to pass in a flag to an hspec test-suite to *disable* `--rerun-all-on-success`, even if it was previously enabled?

I've tried `--no-rerun-all-on-success` and `--rerun-all-on-success=false`, but that just yields command line argument parse failures. (Which is consistent with `spec --help`.)

More specifically I have `--rerun-all-on-success` in my `~/.hspec` file, which I do like as the general default, but I would like to be able to overwrite that occasionally.

sol:


soenkehahn (approved):
Looks good to me. Thanks for the quick turnaround.

sol:
On Hackage

PR #916, opened 2025-04-10
Possible fix for NoListTuplePuns

zlonast:
Replace () with Unit

### DIFF
--- a/hspec-core/package.yaml
+++ b/hspec-core/package.yaml
@@ -23,6 +23,7 @@ dependencies:
   - ansi-terminal >= 0.6.2
   - time
   - transformers >= 0.2.2.0
+  - ghc-prim
   - deepseq
   - HUnit == 1.6.*
   - QuickCheck >= 2.13.1 && < 2.19
--- a/hspec-discover/src/Test/Hspec/Discover/Run.hs
+++ b/hspec-discover/src/Test/Hspec/Discover/Run.hs
@@ -64,6 +64,7 @@ mkSpecModule src conf nodes =
   . showString "{-# OPTIONS_GHC -w -Wall -fno-warn-warnings-deprecations #-}\n"
   . showString ("module " ++ moduleName src conf ++ " where\n")
   . importList nodes
+  . showString "import GHC.Tuple (Unit)\n"
   . showString "import Test.Hspec.Discover\n"
   . maybe driver driverWithFormatter (configFormatter conf)
   . showString "spec :: Spec\n"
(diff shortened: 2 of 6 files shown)


hspec-discover/src/Test/Hspec/Discover/Run.hs
```diff
@@ -64,6 +64,7 @@ mkSpecModule src conf nodes =
   . showString "{-# OPTIONS_GHC -w -Wall -fno-warn-warnings-deprecations #-}\n"
   . showString ("module " ++ moduleName src conf ++ " where\n")
   . importList nodes
+  . showString "import GHC.Tuple (Unit)\n"
```
sol, line 67: This will break things for existing users!

`hspec-discover` can only rely on what is provided by the `hspec` package.

Please instead add this import to `Test.Hspec.Discover` and re-export `Unit` from there.

sol (changes requested):
Hey there 👋

Thanks for your contribution!

1. I left one review comment.  Please address this first and let's see if CI is happy.
2. Once CI is good, I still want to take some time to consider any potential alternatives.

sol:
Hey @zlonast 👋

Do you still want to carry this forward?

zlonast:
@sol Tuples haven't appeared in base yet, so`ghc-experimental` or `ghc-prim` ? :)

hspec-core/package.yaml
```diff
@@ -23,6 +23,7 @@ dependencies:
   - ansi-terminal >= 0.6.2
   - time
   - transformers >= 0.2.2.0
+  - ghc-prim
```
sol, line 26: I think this is not needed

sol (commented):
I think CI is failing.

Maybe we need (2) from https://gitlab.haskell.org/ghc/ghc/-/issues/25952 in GHC:

2. Set the environment variable `GHC_VERSION` to the current GHC version when invoking custom pre-processors, so that pre-processors can emit backwards compatible code without relying on CPP.

We can then start emitting `{-# LANGUAGE ListTuplePuns #-}` if that environment variable is set and the corresponding GHC version understands that extension.

This will not help with GHC 9.10.* / 9.12.* / 9.14.*.  `cabal` could fill the gap and set `GHC_VERSION` when it detects one of these GHC versions (if somebody is gonna look into this, then please also make sure that it takes effect for `cabal exec`).

sol:
https://gitlab.haskell.org/ghc/ghc/-/merge_requests/16224

Issue #914, opened 2025-04-07
timeouts for tests


fisx:
I just made up this thing:

```haskell
timeout :: Int -> Expectation -> Expectation
timeout secs testAction = do
  parentId <- myThreadId
  childId <- forkIO $ do
    threadDelay (secs * 1_000_000)
    killThread parentId
  testAction
  killThread childId

spec = do
  it "..." . timeout 1 $ do
    ...
```

It was easy to write, but maybe it's still something to include into the hspec as a helper?  Let me know if you would like a PR.  If not feel free to reject.  I don't want to bloat this beautiful piece of code.  :)

(Also, we could extend this to work on Specs, not Expectations, so say things like "if this test suite takes longer than 5 sseconds, it should fail"?)

sol:
Hey 👋 

Did you try to implement this in terms of `System.Timeout.timeout`? 

Also, look at https://github.com/hspec/hspec/issues/889#issuecomment-2125837311.

fisx:
> Also, look at [#889 (comment)](https://github.com/hspec/hspec/issues/889#issuecomment-2125837311).

yeah, that's much nicer, thanks!  i'll use that for now, and also make a tiny PR.

PR #918, opened 2025-04-11
Simple timeout combinator.

fisx:
Fixes https://github.com/hspec/hspec/issues/914


fisx:
Not sure this is all according to your taste, but having this would have helped me personally to stay out of your hair.  :)

sol:
@parsonsmatt I think we just need one extension that adds `--timeout=<seconds>` and I guess some timeout combinator `Timeout.apply :: Seconds -> SpeWith a -> SpeWith a` (or maybe `Timeout.seconds :: Seconds -> SpeWith a -> SpeWith a`, not sure).

Either as a separate package or in `-contrib`.

I think that will supersum #889, #914, #918 and #927.

At some point @BebeSparkelSparkel wanted to look into it https://github.com/hspec/hspec/pull/890#issuecomment-2137755272, but from what I understand it's still up for grabs.

PR #924, opened 2025-07-09
docs: write a whole lot more

lf-:
The Example typeclass is core to the usage of hspec, but the
documentation doesn't really explain how it works.

This is my checklist of what I'm trying to fix in the docs:
- What is Example for?
- When you have something in `do` notation, how do you know what
  `Example` type it is? If it looks like IO, is it IO? Should it be
  recommended best practice to use a TypeApplication on `it` to avoid
  error messages from the type inference?
- How do you give an argument to a `SpecWith a` spec to make it a `Spec`
  such that it can actually be executed? What does `Arg` mean in
  practice?

  For example:

  > `Arg a` represents the type that is passed in while executing
  `SpecWith a` instances for function-like types.
- How are you supposed to compose these together? For example, say, you
  have some `Example` type and you want to run an IO action to produce
  some value, run some function, then produce another Example type, how
  do you do that?

  This is related to the usage of libraries like hspec-golden:
  https://hackage.haskell.org/package/hspec-golden-0.2.2.0/docs/Test-Hspec-Golden.html#t:Golden

  ANSWER: you don't. That's why IO Golden implements Example, though, I
  guess. I'm not sure how to put this in the docs so I didn't.
- In the signature:

  `evaluateExample :: e -> Params -> (ActionWith (Arg e) -> IO ()) -> ProgressCallback -> IO Result `

  What is the intended usage of `ActionWith (Arg e)` aka `(Arg e) -> IO ()`
  inside what is effectively `(Arg e) -> IO () -> IO ()`? When is
  something run in the first IO or the second?

  How does this relate to `before`, `after`, `around` and similar?
- How is `ProgressCallback` used? When is it called? What might you use
  it for?
- Is `Params` only intended for QuickCheck?

I remember when I wrote a hspec OpenTelemetry integration a few years
ago I wound up just reading the source code of hspec to figure this out,
but I think it should be documented.


hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -32,16 +32,36 @@ import           Test.Hspec.Core.Tree
...
 
 -- |
 -- @since 2.10.0
 modifyConfig :: (Config -> Config) -> SpecWith a
 modifyConfig f = SpecM $ tell (Endo f, mempty)
 
--- | A writer monad for `SpecTree` forests
+-- | A writer monad for `SpecTree` forests.
+--
+-- This is the monad type passed to `Test.Hspec.Core.Spec.describe` and is used
```
sol, line 60:
```suggestion
-- This is used by `Test.Hspec.Core.Spec.describe` and `Test.Hspec.Core.Spec.it`
```

sol: The type itself is not passed around.  If anything then values of that type are "passed to".

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -32,16 +32,36 @@ import           Test.Hspec.Core.Tree
...
+-- to construct the forest of trees of spec items.
```
sol, line 61:
```suggestion
-- to construct the forest of spec items.
```

A forest always consists of trees. Does it do anything from the perspective of a user to mention this explicitly here?

sol:
> This looks great to me!
> 
> @sol any opposition to me merging this?

@parsonsmatt thanks for taking this on. If you want to shepherd this over the finish line then please go ahead.

hspec-core/src/Test/Hspec/Core/Example.hs
```diff
@@ -40,12 +40,64 @@ import           Test.Hspec.Core.Util
 import           Test.Hspec.Core.QuickCheck.Util (liftHook)
 import           Test.Hspec.Core.Example.Location
 
--- | A type class for examples
+-- | A type class for examples, that is to say, test bodies as used in
+-- `Test.Hspec.Core.Spec.it` and similar functions.
 class Example e where
+  -- | Argument type passed to function-like instances of `Example`.
```
parsonsmatt, line 46: I think this phrasing is a little confusing - it's not about functions so much as "the argument/extra data required to run an `e` as an `Example`"

So if `type Arg X = ()`, that's just a way of saying "I have no requirements"

Similarly we have in codebase `instance Example (SqlPersistT IO ()) where type Arg (SqlPersistT IO ()) = SqlBackend`. Which *is* function-like, in some sense: `SqlPersistT m a ~ ReaderT SqlBackend m a ~ SqlBackend -> m a` - but I think "function-like" makes a reader think more of things like `instance (Example r) => Example (a -> r)` pattern, allowing

```haskell
before mkThing do
  it "has stuff" $ \thing -> do
    thing `shouldBe` "whatever"
```

Possibly:

```suggestion
  -- | Argument type required to run the `Example` test.
```

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -32,16 +32,38 @@ import           Test.Hspec.Core.Tree
...
+-- | A `SpecWith` that can be evaluated directly by the
+-- `Test.Hspec.Core.Runner.hspec` function as it does not require any
+-- parameters.
 type Spec = SpecWith ()
 
+-- | `SpecM` with an empty `ReaderT` environment.
+--
+-- This is the most commonly used spec tree writer type.
+-- It is parameterized over the type @a@, which is the argument type provided
+-- to function-like `Example` instances.
```
parsonsmatt, line 44: I think this is not quite right. The `ReaderT` environment is not empty, the `()` is actually the return value of the `SpecM` monad. From a user perspective, the doc is a little too focused on internals and implementation and not use.

I think I'd call it,

```suggestion
-- | A @'SpecWith' a@ represents a test or group of tests that require an @a@ to evaluate.
-- In the common case, a 'Spec' is a @'SpecWith' ()@, which requires @()@ and can be executed with 'hspec'.
-- 
```

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -79,12 +101,19 @@ mapSpecItem _ = mapSpecItem_
 mapSpecItem_ :: (Item a -> Item b) -> SpecWith a -> SpecWith b
 mapSpecItem_ = mapSpecForest . bimapForest id
 
+-- | Sets the `Params` on the items to be generated by `SpecWith`.
```
parsonsmatt, line 104:
```suggestion
-- | Modifies the `Params` on the items to be generated by `SpecWith`.
```

This is a bit pedantic, feel free to ignore - obviously `const newParams` for `f` here is the "setter" version.

hspec-core/src/Test/Hspec/Core/Spec/Monad.hs
```diff
@@ -79,12 +101,19 @@ mapSpecItem _ = mapSpecItem_
...
 modifyParams :: (Params -> Params) -> SpecWith a -> SpecWith a
 modifyParams f = mapSpecItem_ $ \item -> item {itemExample = \p -> (itemExample item) (f p)}
 
 newtype Env = Env {
+  -- | The path of _parent_ `Hspec.Core.Spec.describe` labels from innermost to
+  -- outermost.
```
parsonsmatt, line 110: this comment would have saved me so much time lol

PR #942, opened 2025-12-13
Revert MicroHs specific changes

sol:
As MicroHs implements more features,  it may be possible to revert some of these changes.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Clock.hs
+++ b/hspec-core/src/Test/Hspec/Core/Clock.hs
@@ -10,10 +10,6 @@ module Test.Hspec.Core.Clock (
 , timeout
 ) where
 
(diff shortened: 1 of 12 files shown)


hspec-core/src/Test/Hspec/Core/Clock.hs
```diff
@@ -37,7 +33,7 @@ toMicroseconds :: Seconds -> Int
 toMicroseconds (Seconds s) = floor (s * 1000000)
 
 getMonotonicTime :: IO Seconds
-#ifdef HAS_GET_MONOTONIC_TIME
+#if MIN_VERSION_base(4,11,0)
```
sol, line 36: https://github.com/augustss/ghc-compat/issues/5

PR #953, opened 2026-02-19
Add a --module-prefix CLI arg to hspec-discover

parsonsmatt:
This PR adds a `--module-prefix` argument to `hspec-discover`, which will allow it to be used in conjunction with `hspec-discover-discover` in subdirectories.


parsonsmatt:
One alternative implementation is to look at `--module-name` and if it has dots, use that as a prefix. ie if I have `--module-name=Foo.Spec` then we imply the prefix of `Foo`.

PR #957, opened 2026-03-05
hspec-discover: Use do-notation instead of `>>`

sol:
This reduces compile times with GHC, but sadly with diminishing returns as the number of spec files increases.

```
125 spec files:

  before: 3.4s
  after:  2.5s

250 spec files:

  before: 9.6s
  after:  5.8s

500 spec files:

  before: 37.1s
  after:  24.6s

1000 spec files:

  before: 163.7s
  after:  147.8s
```

Still, as this is a strict improvement and makes the generated test driver easier to read, I think this is a good change.

In addition, from other things I tried, I suspect that this may positively impact `-O0`.

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Spec.hs
+++ b/hspec-core/src/Test/Hspec/Core/Spec.hs
@@ -106,6 +106,7 @@ describe :: HasCallStack => String -> SpecWith a -> SpecWith a
 describe label = withEnv pushLabel . mapSpecForest (return . specGroup label)
   where
     pushLabel (Env labels) = Env $ label : labels
+{-# OPAQUE describe #-}
 
 -- | @context@ is an alias for `describe`.
 context :: HasCallStack => String -> SpecWith a -> SpecWith a
--- a/hspec-discover/src/Test/Hspec/Discover/Run.hs
+++ b/hspec-discover/src/Test/Hspec/Discover/Run.hs
@@ -116,20 +116,22 @@ moduleNames = fromForest
       Spec name -> [name ++ "Spec"]
       Hook name forest -> name : fromForest forest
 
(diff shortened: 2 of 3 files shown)


sol:
More data:

```
1000 spec files:

  before: 163.7s
  after:  147.8s

  with -O0: ~70s
  in ghci:   ~5s
```

Most of the time is spent in the simplifier (and without `-O0` an equal amount is spent during code generation).

In GHCi the lion's share, ~3.5s, is spent on renaming / type checking.

Right now I suspect that this may be due to overeager inlining and that explicitly marking `describe` as `NOINLINE` might solve the issue.

sol:
Apparently, for some reason, it makes a difference wether I give `-O0` via an `OPTIONS_GHC` pragma, or via `ghc-options` in `.cabal`. 

If I do it via a pragma, compile time for the test driver goes down from ~140s to ~70s.

If I do it in `.cabal` then compile time for the test driver goes down to ~3.5s.

In the later case each individual spec module is also compiled with `-O0`.  Could that possibly matter? 🤷

sol:
The real reason for the difference between `-O0` in a pragma vs globally is that in the later case `-fignore-interface-pragmas` is enabled before Interface files for third-party dependencies are  read.

With `-fignore-interface-pragmas` GHC strips any unfoldings from interface files when reading them.  In particular, this matters for `transformers`.

Without `-fignore-interface-pragmas`, once unfoldings are available GHC apparently more or less applies them mindlessly, regardless of the optimization level.  This can be confirmed by looking at the output of `-ddump-rule-firings`.

Using `sequence_` instead of `do`-notation offers a workaround.

sol:
Closing in favor of #958.

Issue #952, opened 2026-02-18
`hspec-discover` should stop recursively searching a directory if it finds a `Spec.hs` module


parsonsmatt:
Our `test/Spec.hs` module takes around 150s to compile on average. We have a shitload of `test/**/*Spec.hs` modules. One way we can ameliorate this problem is by stuffing `test/Foo/Spec.hs` modules which call `hspec-discover` themselves, and doing this for each subdirectory of `test/`. These would each be significantly smaller than the main `Spec.hs` file and reduce total compilation.

To support his, `hspec-discover` should have a "stop recursing" switch. Possibly just detecting a module named `Spec.hs` in the directory. If it sees that, it only imports `Spec.hs` and plugs in `Foo.Spec.spec` instead of recursing into that directory tree.

sol:
> That compilation time is sub/super-linear

Yes, exactly.

If it were linear then a spec driver with 260 imports should take 15s to compile, and a spec driver with 26 imports 1.5s.  Not impossible, but more than what I would have expected.

> in count of modules imported?

The number of imports themselves is not my prime suspect. But I think this can be narrowed down further, by disabling optimizations, or code generation altogether.

parsonsmatt:
We are already compiling the module with `-O0`. GHC is known to be superlinear in a lot of things, so I kinda suspect we're just looking at about as much Stuff as GHC can handle. Breaking things up into separate modules has several other benefits, as well - each `TestGroup` module gets compiled as soon as all of the tests it discovers are done, which speeds things up significantly.

I wouldn't be surprised if there was some suboptimal GHC behavior that contributed to this slowdown, but it's solidly in "edge case" territory - a single expression with 2.6k `>>` calls! 

TBH I'm not particularly interested in generating the timing data on this. At some point in the future, we'll be using a buck2 based test harness which will likely remove `hspec-discover` from our main test suite.

sol:
```
360 spec files: 18.7s
180 spec files: 5.7s

with `-O0`

360 spec files: 8.4s
180 spec files: 3.1s

do-notation instead of >>

360 spec files: 5.9s
180 spec files: 2.4s

(all times include linking)
```

sol:
> We are already compiling the module with `-O0`.

@parsonsmatt Where exactly do you put the `-O0`? 

Can you try compiling the entire test suite with `-fignore-interface-pragmas` and possibly also `-fomit-interface-pragmas`?

sol:
[...]

PR #958, opened 2026-03-06
hspec-discover: Prevent GHC from overeagerly simplifying the generated code

sol:


hspec-discover/test/Test/Hspec/Discover/RunSpec.hs
```diff
@@ -48,10 +48,10 @@ spec = do
...
         , "import Test.Hspec.Discover"
         , "spec :: Spec"
-        , "spec = " ++ unwords [
-               "describe \"Foo\" FooSpec.spec"
-          , ">> describe \"Foo.Bar\" Foo.BarSpec.spec"
-          , ">> describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec"
+        , "spec = sequence_ [" ++ unwords [
+            "describe \"Foo\" FooSpec.spec,"
+          , "describe \"Foo.Bar\" Foo.BarSpec.spec,"
+          , "describe \"Foo.Bar.Baz\" Foo.Bar.BazSpec.spec]"
```
parsonsmatt, line 54: interestingly enough, I remember having some major compile-time performance boons by switchig code from generating `<>` chains to using `mconcat` - I wonder if this is something that is generalizable...

parsonsmatt (approved):
hell yeah

parsonsmatt (commented):
I've tested this out on our codebase and ran several builds with the patch. Using our `opentelemetry-plugin` to get per-module level timing data, unfortunately, the `hspec-discover` enabled modules did not observe any speedup. 

Image

This graph shows builds of our `TestGroup` modules over the last 7 days. The spike on the far right shows builds now, with the patch enabled. Build times are slightly slower than without the patch. That's *pretty surprising*, but if you are observing improvements locally, then I wouldn't want to block on this.

One reason why I wouldn't block on this: this is with ghci and we have a heavily patched GHC 9.10.1 so it's entirely possile that we're tripping some regression that we introduced *or* there's some fix that we don't have incorporated.

sol:
@parsonsmatt sorry, this is not useful to me.

1. You originally reported build times of ~150s for a top level test driver with 2.6k.
2. Locally, on my machine, from what I remember, a minimal example with 2.6K imports took even longer than 150s (don't quote me on the exact number, but I think it was around ~180s, maybe even longer).
3. This PR brings the build time for a test driver with 6k imports (not 2.6k!) down to 7s.

Given this, my expectation is that this PR, applied to your original situation, will bring build times down from 150s to somewhere in the ~2s range.

@parsonsmatt are you willing to try this PR in isolation and confirm (or disproof) my expectation?

Once we properly understand the impact of this PR, we can discuss any further steps (e.g. #954 or `-O0`).

But please look at this PR in isolation (without #954), because from my experience mashing things together rarely gives you a clear picture of what's actually happening.

> this is with ghci

What is with `ghci`?  Your originally reported issue (-150s build times for `Spec.hs` with 2.6k imports) should never have been an issue with `ghci`.  In `ghci` "build" (or rather "interpretation") times should always have been close to instant, or at most in the low single digit seconds range.  From everything I tried and understand, GHC simply doesn't apply the unfolding that are responsible for #952 in `--interactive` mode.

parsonsmatt:
> @parsonsmatt are you willing to try this PR in isolation and confirm (or disproof) my expectation?

I've already migrated our codebase over to `hspec-discover-discover` in tandem with my PR. If I want to revert that, I need to check out a several month old version of our codebase, or undo the migration. In either case, rebuilding our nix shell with a new `hspec-discover` takes several hours, during which time I cannot do any other computation intensive work. If I revert entirely back to when I initially filed the report, I'll have to rebuild GHC also, which makes it a daylong process to even start building. What you're asking isn't unreasonable, and if the cost weren't so great, I'd gladly do it.

Your own data suggests that your PR is a nice improvement. If I were you, I'd just merge it. We aren't observing any benefit per my prior data. *But*, that could just come down to our patched GHC version having some different behavior, so our data may not be representative anyway.

The times that I'm looking at are with GHCi. The *really* long times I reported earlier are from a `cabal` based build. GHCi is faster, but the modules still take quite some time to load in GHCi. That data is out of our retention window, so I can't bring it up any longer.

PR #962, opened 2026-05-14
Self-deregister completed jobs from the JobQueue cancel list

### LINKED ISSUE #961: `CancelQueue` holds on to memory for the entire process run
The work test suite recently experienced memory pressure which caused it to fail in CI. Unfortunately, we are only able to run `+RTS -hT` profiling. Due to weirdness in our nix build, a proper profiling run isn't feasible. In my investigation, I found that the we were accumulating a lot of `:` constructors and `STACK`. Digging around for long running lists, I found `CancelQueue`, which accumulates an `[Async ()]` representing each of the test cases.  It seems like the `Async ()` kept in memory was causing GHC to retain the thread and the thread's `STACK`.

I replaced the `[Async ()]` with a `Map ThreadId (Async ())` and introduced a deregistration-on-finalization to promptly cancel the asyncs. With this patch, `STACK` is stable at around 44MB in the `-hT` profile, compared to ~600MB (and steadily growing) prior to the patch. I did not get a complete test run for either case: the tests just take too long on my machine, and the trend was present. However, even a partial run indicates about 60% memory savings on our test run.

I'll open a PR shortly for the `Map ThreadId (Async ())` fix.

I made a synthetic test case for this, with a bunch of threads doing `pure ()`. This demonstrated the problem with the growing `:` constructor, but did not demonstrate the `STACK` gain - we'd need tests to do some legitimate real work in the `STACK` to do that. The synthetic case showed benefit, but it was substantially understated.

The #962 PR includes a regression test- it is copied here.

```haskell
-- Regression test for hspec#961:
--   JobQueue.cancelQueue retained one Async (and its underlying TSO) per spec
--   item for the lifetime of the run, growing O(n) in the number of test
--   cases. After the fix, completed workers self-deregister, so the queue
--   only holds in-flight jobs.
--
-- Reproducer: a synthetic suite with nested describes (similar shape to a
-- real test tree) where each leaf does a small amount of work. Run with
-- `+RTS -hT -RTS` to produce a heap profile by closure type; the
-- accompanying run.sh inspects Spec.hp to validate that retention of
-- worker-owned closure types stays roughly constant in the suite size.

module Main (main) where

import           Control.Monad (forM_)
import           Test.Hspec

outer, middle, leaf :: Int
outer = 50
middle = 20
leaf = 50

main :: IO ()
main = hspec $ describe "issue-961" $
  forM_ [1 .. outer] $ \ i ->
    describe ("group-" <> show i) $
      forM_ [1 .. middle] $ \ j ->
        describe ("subgroup-" <> show j) $
          forM_ [1 .. leaf] $ \ k ->
            it ("test-" <> show i <> "-" <> show j <> "-" <> show k) $ do
              let xs = map (\ x -> x * x + i + j + k) [1 .. 200 :: Int]
              sum xs `shouldBe` sum xs
```

To run,

```
#!/bin/bash
#
# Regression test for hspec#961.
#
# Runs the synthetic 50k-test suite under `+RTS -hT -RTS` to produce a heap
# profile by closure type. Then inspects the last non-empty sample (which
# captures retention right before process exit) and validates that the bytes
# held in TSO + STACK + Async closures — i.e. the storage the cancel queue
# would retain per worker — stays roughly constant in the suite size.
#
# On the broken JobQueue, ~50k worker TSOs/STACKs are kept alive through the
# end of the run; STACK alone reaches ~45 MB. With the fix, completed workers
# self-deregister and the same closures stay under ~5 MB total.

set -o errexit
cd `dirname "$0"`

THRESHOLD_BYTES=$((15 * 1024 * 1024))

ghc -threaded -rtsopts -fforce-recomp Spec.hs
./Spec -f silent +RTS -hT -i0.05 -RTS

awk -v threshold="$THRESHOLD_BYTES" '
  /^BEGIN_SAMPLE/ { time=$2; tso=0; stack=0; async=0; have=0 }
  /^TSO[[:space:]]/                              { tso=$NF;   have=1 }
  /^STACK[[:space:]]/                            { stack=$NF; have=1 }
  /Control\.Concurrent\.Async\.Async[[:space:]]/ { async=$NF; have=1 }
  /^END_SAMPLE/ {
    if (have) {
      last_time = time
      last_tso = tso
      last_stack = stack
      last_async = async
    }
  }
  END {
    total = last_tso + last_stack + last_async
    printf "issue-961: last non-empty sample at %s\n", last_time
    printf "issue-961:   TSO    = %d bytes\n", last_tso
    printf "issue-961:   STACK  = %d bytes\n", last_stack
    printf "issue-961:   Async  = %d bytes\n", last_async
    printf "issue-961:   total  = %d bytes (threshold %d)\n", total, threshold
    if (total >= threshold) {
      print "issue-961: FAIL — cancel queue retained completed workers"
      exit 1
    }
    print "issue-961: PASS"
  }
' Spec.hp

rm Spec.o Spec.hi Spec Spec.hp
```

parsonsmatt:
The cancel queue (`IORef [Async ()]`) accumulated every spawned job for the duration of a run. Completed asyncs were never removed, so a 50k test run retained ~48 MB of `Async` references plus another ~7 MB of RTS thread bookkeeping that GHC could not reclaim while ThreadIds remained reachable.

Switch to `IORef (Map ThreadId (Async ()))` and have each worker remove its own entry in a `finally`. A small `ready` MVar closes the registration race between `async worker` (which begins running immediately) and the parent's insert. `mask_` around spawn+register preserves the existing exception semantics: an async exception cannot strand a registered-but-ungated worker.

Steady-state cancel queue size is now bounded by concurrency (typically `nproc`) rather than total job count. Measured on a synthetic 50k-spec suite: end-of-run heap drops from ~70 MB to ~14 MB. On my work codebase, this PR results in 64% less memory use on a partial run, and the delta is growing linearly.

Fixes #961

### DIFF
--- a/hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
@@ -15,7 +15,9 @@ import           Prelude ()
 import           Test.Hspec.Core.Compat
 
 import           Control.Concurrent
-import           Control.Concurrent.Async (Async, async, waitCatch, cancelMany)
+import           Control.Concurrent.Async (Async, async, asyncThreadId, cancel, waitCatch, cancelMany)
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
 
 import           Control.Monad.IO.Class
 
@@ -33,16 +35,16 @@ data Semaphore = Semaphore {
 , _signal :: IO ()
 }
 
-type CancelQueue = IORef [Async ()]
+type CancelQueue = IORef (Map ThreadId (Async ()))
 
 withJobQueue :: Int -> (JobQueue -> IO a) -> IO a
 withJobQueue concurrency = bracket new cancelAll
   where
     new :: IO JobQueue
-    new = JobQueue <$> newSemaphore concurrency <*> newIORef []
+    new = JobQueue <$> newSemaphore concurrency <*> newIORef Map.empty
 
     cancelAll :: JobQueue -> IO ()
-    cancelAll (JobQueue _ cancelQueue) = readIORef cancelQueue >>= cancelMany
+    cancelAll (JobQueue _ cancelQueue) = readIORef cancelQueue >>= cancelMany . Map.elems
 
 newSemaphore :: Int -> IO Semaphore
 newSemaphore capacity = do
@@ -72,10 +74,20 @@ data Partial progress a = Partial progress | Done
 runConcurrently :: forall m progress a. MonadIO m => Semaphore -> CancelQueue -> Job IO progress a -> IO (Job m progress (Either SomeException a))
 runConcurrently (Semaphore wait signal) cancelQueue action = do
   result :: MVar (Partial progress a) <- newEmptyMVar
+  ready :: MVar () <- newEmptyMVar
   let
+    awaitWorkerStart :: IO ()
+    awaitWorkerStart = readMVar ready
+
+    signalWorkerStart :: IO ()
+    signalWorkerStart = putMVar ready ()
+
     worker :: IO a
-    worker = bracket_ wait signal $ do
-      interruptible (action partialResult) `finally` done
+    worker = (do
+        awaitWorkerStart
+        bracket_ wait signal $
+          interruptible (action partialResult) `finally` done
+      ) `finally` deregisterSelf
       where
         partialResult :: progress -> IO ()
         partialResult = replaceMVar result . Partial
@@ -83,10 +95,27 @@ runConcurrently (Semaphore wait signal) cancelQueue action = do
         done :: IO ()
         done = replaceMVar result Done
 
-    pushOnCancelQueue :: Async a -> IO ()
-    pushOnCancelQueue = (modifyIORef cancelQueue . (:) . void)
+        deregisterSelf :: IO ()
+        deregisterSelf = myThreadId >>= deregister
 
-  job <- bracket (async worker) pushOnCancelQueue return
+    modifyPending :: (Map ThreadId (Async ()) -> Map ThreadId (Async ())) -> IO ()
+    modifyPending f =
+      atomicModifyIORef' cancelQueue $ \ pending -> (f pending, ())
+
+    register :: Async () -> IO ThreadId
+    register workerAsync = do
+      let tid = asyncThreadId workerAsync
+      modifyPending (Map.insert tid workerAsync)
+      return tid
+
+    deregister :: ThreadId -> IO ()
+    deregister tid = modifyPending (Map.delete tid)
+
+  job <- mask_ $
+    bracketOnError (async worker) cancel $ \ workerAsync ->
+      bracketOnError (register (void workerAsync)) deregister $ \ _tid -> do
+        signalWorkerStart
+        return workerAsync
 
   let
     waitForResult :: (progress -> m ()) -> m (Either SomeException a)
--- a/hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
+++ b/hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
@@ -15,7 +15,7 @@ import           Prelude ()
 import           Test.Hspec.Core.Compat
 
 import           Control.Concurrent
-import           Control.Concurrent.Async (Async, async, asyncThreadId, cancel, waitCatch, cancelMany)
+import           Control.Concurrent.Async (Async, async, asyncThreadId, cancel, cancelMany)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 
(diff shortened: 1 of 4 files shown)


hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
```diff
@@ -39,16 +41,16 @@ data Semaphore = Semaphore {
 , _signal :: IO ()
 }
 
-type CancelQueue = IORef [Async ()]
+type CancelQueue = IORef (Map ThreadId (Async ()))
```
parsonsmatt, line 38: This is the basis of the first fix in this PR. Instead of accumulating all of the test run `Async ()` into a list, we keep track of a `Map ThreadId (Async ())`. Tests deregister themselves from this when they complete, which keeps this from growing without bound.

hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
```diff
@@ -73,34 +75,57 @@ runSequentially cancelQueue action = do
   job <- runConcurrently (Semaphore wait pass) cancelQueue action
   return $ \ notifyPartial -> signal >> job notifyPartial
 
-data Partial progress a = Partial progress | Done
+data Partial progress a = Partial progress | Done (Either SomeException a)
```
parsonsmatt, line 72: This is the basis of the *second* fix in this PR. The synthetic benchmark test suite exposed a problem that my work test suite did not- the `Async` were still being held onto, and the heap was still accumulating! The issue was doing `waitCatch` on the async- this was holding on to that memory. By allowing the result to be communicated entirely through the `MVar`, this allowed us to release the reference to the `Async` earlier.

hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
```diff
@@ -73,34 +75,57 @@ runSequentially cancelQueue action = do
...
 
 runConcurrently :: forall m progress a. MonadIO m => Semaphore -> CancelQueue -> Job IO progress a -> IO (Job m progress (Either SomeException a))
 runConcurrently (Semaphore wait signal) cancelQueue action = do
   result :: MVar (Partial progress a) <- newEmptyMVar
+  ready :: MVar () <- newEmptyMVar
```
parsonsmatt, line 77: New `ready` `MVar` is necessary in order to signal "your finalizer has been installed"

hspec-core/src/Test/Hspec/Core/Runner/JobQueue.hs
```diff
@@ -73,34 +75,57 @@ runSequentially cancelQueue action = do
...
+      modifyPending (Map.insert tid workerAsync)
+      return tid
+
+    deregister :: ThreadId -> IO ()
+    deregister tid = modifyPending (Map.delete tid)
+
+  mask_ $
+    bracketOnError (async worker) cancel $ \ workerAsync ->
+      bracketOnError (register workerAsync) deregister $ \ _tid ->
+        signalWorkerStart
```
parsonsmatt, line 114: The `workerAsync` variable does not escape this scope, so GC can collect it as soon as it writes to the `result` var and exits.

parsonsmatt (commented):
I believe this is ready for review.

I'm appreciative for the synthetic benchmark request - this did demonstrate a problem. However, curiously, the first fix (`type CancelQueue = IORef (Map ThreadId (Async ()))`) did not reliably pass the included regression test! This surfaced the *second* fix (`Done (Either SomeException a)`, which now shows pretty flat results in the regression test in this PR. 

What's strange is that the `CancelQueue` change resulted in a *huge* memory reduction in our test suite - 60%, but the `Done (Either SomeException a)` change only improved things by another 5% or so. It wouldn't surprise me if the regression test did not properly trigger the behavior that caused the `CancelQueue` to be so problematic in our test suite. Our test suite setup is quite complicated and it would be extremely difficult to do a proper reduction that demonstrated the `CancelQueue` issue.

sol:
@parsonsmatt this PR increases total residency for me with GHC 9.14.1 (and 9.12.4) and my repro from #972.

**this PR**
```console
$ cabal build && $(cabal list-bin example) --format progress +RTS -s
...
Finished in 5.3110 seconds
50000 examples, 0 failures
...
             319 MiB total memory in use (0 MiB lost due to fragmentation)
...
```

**vs `main`**
```console
$ cabal build && $(cabal list-bin example) --format progress +RTS -s
...
Finished in 5.9424 seconds
50000 examples, 0 failures
...
             214 MiB total memory in use (0 MiB lost due to fragmentation)
...
```
**vs #970**
```console
$ cabal build && $(cabal list-bin example) --format progress +RTS -s
...
Finished in 4.0329 seconds
50000 examples, 0 failures
...
             124 MiB total memory in use (0 MiB lost due to fragmentation)
...
```

I think #970 is the way forward.  Do you want to close this?

PR #963, opened 2026-05-14
Demonstration: issue-961 regression test alone (no fix)

parsonsmatt:
**Not for merge.** This PR contains only the issue-961 regression test (from #962) without the JobQueue fix, so we can observe CI behavior against unmodified hspec and confirm the test discriminates the regression.

Expected outcome: the `Run issue-961 regression test` step should fail on every GHC matrix row that runs it (GHC 8.8 through 9.12) — proving the test catches the cancel-queue leak in stock hspec.

Companion to #962, which contains both the fix and this regression test.


sol:
Closing in favor of #970.

